PR #7, opened 2021-11-09
[WIP][Feature] Server excluded bad disk

jerqi:


server/src/main/java/com/tencent/rss/server/ShuffleFlushManager.java
```diff
@@ -145,6 +150,7 @@ public void addToFlushQueue(ShuffleDataFlushEvent event) {
 
   private void flushToFile(ShuffleDataFlushEvent event) {
 
+    event.pickStorageIndex(storageBasePaths.length);
```
colinmjj, line 153: If storage is broken during the shuffle, shuffle data will be located in different location for the same partition?

jerqi: Yes, partition will locate an another storage. Actually, once a storage is broken,  I hope shuffle will fail fast. I hope to implement this function.

duanmeng:
Could you write a description about your strategy for juding whether a disk is damaged or repaired and the following actions such as the recovery procedure. And what is this pr's goal.

server/src/main/java/com/tencent/rss/server/ShuffleFlushManager.java
```diff
@@ -156,86 +162,98 @@ private void flushToFile(ShuffleDataFlushEvent event) {
...
+  }
+
+  private void processWriteResult(ShuffleDataFlushEvent event, long duration, boolean writeSuccess) {
+    if (shuffleServer != null) {
+      shuffleServer.getShuffleBufferManager().releaseMemory(event.getSize(), true, false);
+      storageErrorMap.putIfAbsent(event.getStorageIndex(), new AtomicInteger(0));
+      AtomicInteger errors = storageErrorMap.get(event.getStorageIndex());
+      if (writeSuccess) {
+        errors.updateAndGet(x -> {
+          --x;
```
duanmeng, line 212: `errors.updateAndGet(x -> --x < 0 ? 0 : x)`?

server/src/main/java/com/tencent/rss/server/ShuffleFlushManager.java
```diff
@@ -156,86 +162,98 @@ private void flushToFile(ShuffleDataFlushEvent event) {
...
+          if (x < 0) {
+            return 0;
+          }
+          return x;
+        });
+        LOG.debug("Flush to file success in " + duration + " ms and release " + event.getSize() + " bytes");
```
duanmeng, line 218: use log placeholder

server/src/main/java/com/tencent/rss/server/ShuffleFlushManager.java
```diff
@@ -156,86 +162,98 @@ private void flushToFile(ShuffleDataFlushEvent event) {
...
+      } else {
+        if (storageBasePaths.length > 1 && errors.incrementAndGet() > excludeStorageThreshold) {
+          ShuffleStorageUtils.addExcludeStorage(event.getStorageIndex());
+          LOG.error("Add {} to exclude storage set", storageBasePaths[event.getStorageIndex()]);
         }
+        LOG.error("Flush to file for " + event + " failed in "
```
duanmeng, line 224: use log placeholder

jerqi:
> Could you write a description about your strategy for juding whether a disk is damaged or repaired and the following actions such as the recovery procedure. And what is this pr's goal.

No repairing operation. I will update pr's description according to template.

server/src/main/java/com/tencent/rss/server/ShuffleFlushManager.java
jerqi: OK

server/src/main/java/com/tencent/rss/server/ShuffleFlushManager.java
jerqi: OK

server/src/main/java/com/tencent/rss/server/ShuffleFlushManager.java
jerqi: I prefer the current implement. Because it's more easy to understand.

jerqi:
In our production environment, we find  that raid disks can provide the fault tolerence and better performance. So the pr seems not to be much important.

PR #8, opened 2021-11-09
[Improvement] Replace runtime exception with rss related exception in…

colinmjj:
**What changes were proposed in this pull request?**
This PR add RSS related exception which can be identify there has some problem in shuffle server 

**Why are the changes needed?**
Improve the exception

**Does this PR introduce any user-facing change?**
No

**How was this patch tested?**
base on current UT


client-spark/common/src/main/java/org/apache/spark/shuffle/RssCausedException.java
```diff
@@ -0,0 +1,28 @@
...
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+
+package org.apache.spark.shuffle;
+
+public class RssCausedException extends RuntimeException {
+
+  public RssCausedException(String message) {
```
jerqi, line 24: Spark cannot recognize the RssCauseException，should we put some specific string to error message？

jerqi:
We replace runtimeException in class ShuffleGrpcClient with RssCausedException

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
```diff
@@ -217,10 +219,12 @@ private void requestExecutorMemory(long leastMem) {
...
         String message = "Can't get memory to cache shuffle data, request[" + askExecutorMemory
-            + "], got[" + gotMem + "]," + " WriteBufferManager allocated[" + allocatedBytes + "] task used[" + used
-            + "], consider to optimize 'spark.executor.memory'," + " 'spark.rss.writer.buffer.spill.size'.";
+            + "], got[" + gotMem + "]," + " WriteBufferManager allocated[" + allocatedBytes
+            + "] task used[" + used + "]. It may be caused by shuffle server is full of data"
+            + " or consider to optimize 'spark.executor.memory',"
+            + " 'spark.rss.writer.buffer.spill.size'.";
         LOG.error(message);
-        throw new OutOfMemoryError(message);
+        throw new RssCausedException(message);
```
duanmeng, line 227: Should we make the exception classification more fine-grained, such as `RssCommitException`, `RssSendDataException`, etc.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssCausedException.java
colinmjj: With such rss related exception,  it can be got from spark event listener.

colinmjj:
> We replace runtimeException in class ShuffleGrpcClient with RssCausedException

Exception in ShuffleGrpcClient means network problem or shufffle server problem. 
The pr is target to throw RssCausedException and can be seen by spark directly, it will help event listener to catch it easily

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
colinmjj: Yes, but current pr is target to throw RssCausedException and can be seen by spark directly, it will help event listener to catch it easily. The fine-grained exception can be created if necessary.

jerqi:
> > We replace runtimeException in class ShuffleGrpcClient with RssCausedException
> 
> Exception in ShuffleGrpcClient means network problem or shufffle server problem. The pr is target to throw RssCausedException and can be seen by spark directly, it will help event listener to catch it easily

But these exceptions are also caused by rss.

client-spark/common/src/main/java/org/apache/spark/shuffle/RssCausedException.java
jerqi: Is there spark listener Code?

colinmjj:
> > > We replace runtimeException in class ShuffleGrpcClient with RssCausedException
> > 
> > 
> > Exception in ShuffleGrpcClient means network problem or shufffle server problem. The pr is target to throw RssCausedException and can be seen by spark directly, it will help event listener to catch it easily
> 
> But these exceptions are also caused by rss.

Exception in ShuffleGrpcClient will be wrapped, in such case, RssCausedException maybe the child of RssCausedException

client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
duanmeng: OK

colinmjj:
> > > We replace runtimeException in class ShuffleGrpcClient with RssCausedException
> > 
> > 
> > Exception in ShuffleGrpcClient means network problem or shufffle server problem. The pr is target to throw RssCausedException and can be seen by spark directly, it will help event listener to catch it easily
> 
> But these exceptions are also caused by rss.

ok, do the replacement

colinmjj:
@jerqi update pr according to comments

jerqi:
ShuffleWriteClientImpl also need to be replaced.

colinmjj:
> ShuffleWriteClientImpl also need to be replaced.

done

jerqi (approved):

PR #12, opened 2021-11-10
Fix Dependabot alerts

duanmeng:
Fix github dependabot alerts

Dependabot alerts should be fixed

No

UT and integration test

### DIFF
--- a/integration-test/spark3/pom.xml
+++ b/integration-test/spark3/pom.xml
@@ -91,11 +91,7 @@
             <type>test-jar</type>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <version>4.11</version>
-        </dependency>
+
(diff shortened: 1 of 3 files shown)


colinmjj (approved):
LGTM

integration-test/spark3/pom.xml
```diff
@@ -112,26 +108,17 @@
...
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-core</artifactId>
-            <version>2.10.0</version>
             <scope>test</scope>
         </dependency>
 
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-minicluster</artifactId>
-            <scope>test</scope>
```
jerqi, line 128: why do we delete scope?

duanmeng: Its scope is set in the central dependency management in the parent pom.

PR #6, opened 2021-11-09
[Feature] Read index file first in local mode

duanmeng:
### What changes were proposed in this pull request?
1. Add read shuffle index api on both server and client side
2. Read all index data from server and parse it in client side
3. Use offset and length to read data directly from server side

### Why are the changes needed?
The current read procedure of Firestorm's local mode is that the client send data fetch request to the server. The server parse index file and maintain index file segements divide by *readBufferSize* and it read and parse each segment once it received a data fetch request from client. 
We want to change the current read procedure, client fetch index file from server directly, parse it to segments according to *readBufferSize* and then fetch data by offset and length. We can fetch shuffle data precisely and it's convenient to  realize server HA and read replica on demand.
A index segment has 40b and a block is at lease 3kb (the worst case, normally it should be 1-32mb), so the index data's magnitude is MB level at most to fit into memory in client side.

### Does this PR introduce _any_ user-facing change?
No.

### How was this patch tested?
Integration/unit test and terasort/tpc-ds test to verify the result and performance.

PR #21, opened 2021-11-22
[Feature] Read index file first in hdfs mode

duanmeng:

1. Add read shuffle index api in hdfs read handler
2. Read all index data from hdfs
3. Use offset and length to read data from hdfs

As mentioned in #6, we want to read index first and use it to read data and
then we could realize server HA and read replica on demand.

No

UT, integration test, TPC-DS and terasort.

### DIFF
--- a/storage/src/main/java/com/tencent/rss/storage/api/ShuffleReader.java
+++ b/storage/src/main/java/com/tencent/rss/storage/api/ShuffleReader.java
@@ -26,5 +26,7 @@ public interface ShuffleReader {
 
   byte[] readData(long offset, int length);
 
+  byte[] readIndex();
+
   List<FileBasedShuffleSegment> readIndex(int limit) throws IOException, IllegalStateException;
 }
--- a/storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
+++ b/storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
@@ -18,8 +18,18 @@
 
 package com.tencent.rss.storage.handler.impl;
 
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Lists;
+import com.tencent.rss.common.ShuffleDataResult;
+import com.tencent.rss.common.ShuffleDataSegment;
+import com.tencent.rss.common.ShuffleIndexResult;
+import com.tencent.rss.common.util.RssUtils;
 import com.tencent.rss.storage.util.ShuffleStorageUtils;
 
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map.Entry;
+import java.util.stream.Collectors;
 import org.apache.hadoop.conf.Configuration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -27,6 +37,12 @@ import org.slf4j.LoggerFactory;
 public class HdfsClientReadHandler extends AbstractHdfsClientReadHandler {
 
   private static final Logger LOG = LoggerFactory.getLogger(HdfsClientReadHandler.class);
+
+  private final List<ShuffleDataSegment> shuffleDataSegments = Lists.newArrayList();
+  // TODO: Refactor read handler to wrapper a single index file and its indexed data file into a handler.
+  private final Iterator<Entry<String, HdfsFileReader>> readerIterator;
+  private HdfsFileReader currentIndexReader;
+  private HdfsFileReader currentDataReader;
 
   public HdfsClientReadHandler(
       String appId,
@@ -51,7 +67,92 @@ public class HdfsClientReadHandler extends AbstractHdfsClientReadHandler {
         ShuffleStorageUtils.getShuffleDataPathWithRange(appId,
             shuffleId, partitionId, partitionNumPerRange, partitionNum));
     init(fullShufflePath);
-    readAllIndexSegments();
+    readerIterator = indexReaderMap
+        .entrySet()
+        .stream()
+        .sorted(Entry.comparingByKey())
+        .collect(Collectors.toList())
+        .iterator();
   }
 
+  public ShuffleIndexResult readShuffleIndex() {
+    long start = System.currentTimeMillis();
+    try {
+      byte[] indexData = currentIndexReader.readIndex();
+      LOG.info("Read index files for shuffleId[{}], partitionId[{}] for {} ms",
+          shuffleId, partitionId, System.currentTimeMillis() - start);
+      return new ShuffleIndexResult(indexData);
+    } catch (Exception e) {
+      LOG.info("Fail to read index files for shuffleId[{}], partitionId[{}]", shuffleId, partitionId);
+    }
+
+    return new ShuffleIndexResult();
+  }
+
+  public ShuffleDataResult readShuffleData(ShuffleDataSegment shuffleDataSegment) {
+    int expectedLength = shuffleDataSegment.getLength();
+    if (expectedLength < 0) {
+      LOG.error("Invalid segment {}", shuffleDataSegment);
+      return null;
+    }
+
+    byte[] data = currentDataReader.readData(shuffleDataSegment.getOffset(), expectedLength);
+    if (data.length != expectedLength) {
+      LOG.error("Fail to read expected[{}] data, actual[{}] and segment is {}",
+          expectedLength, data.length, shuffleDataSegment);
+      return null;
+    }
+
+    return new ShuffleDataResult(data, shuffleDataSegment.getBufferSegments());
+  }
+
+  @Override
+  public ShuffleDataResult readShuffleData(int segmentIndex) {
+    if (segmentIndex >= shuffleDataSegments.size()) {
+      ShuffleIndexResult shuffleIndexResult = null;
+      while (readerIterator.hasNext()) {
+        Entry<String, HdfsFileReader> entry = readerIterator.next();
+        String key = entry.getKey();
+        currentIndexReader = entry.getValue();
+        if (currentIndexReader == null) {
+          LOG.error("Index reader of {} is invalid!", key);
+          continue;
+        }
+
+        currentDataReader = dataReaderMap.get(entry.getKey());
+        if (currentDataReader == null) {
+          LOG.error("Data reader of {} is invalid!", key);
+          continue;
+        }
+
+        shuffleIndexResult = readShuffleIndex();
+        if (!shuffleIndexResult.isEmpty()) {
+          break;
+        }
+      }
+
+      if (shuffleIndexResult == null || shuffleIndexResult.isEmpty()) {
+        return null;
+      }
+
+      List<ShuffleDataSegment> cur = RssUtils.transIndexDataToSegments(shuffleIndexResult, readBufferSize);
+      shuffleDataSegments.addAll(cur);
+    }
+
+    if (segmentIndex >= shuffleDataSegments.size()) {
+      return null;
+    }
+
+    return readShuffleData(shuffleDataSegments.get(segmentIndex));
+  }
+
+  @VisibleForTesting
+  List<ShuffleDataSegment> getShuffleDataSegments() {
+    return shuffleDataSegments;
+  }
+
+  @VisibleForTesting
+  public Iterator<Entry<String, HdfsFileReader>> getReaderIterator() {
+    return readerIterator;
+  }
 }
--- a/storage/src/main/java/com/tencent/rss/storage/handler/impl/AbstractHdfsClientReadHandler.java
+++ b/storage/src/main/java/com/tencent/rss/storage/handler/impl/AbstractHdfsClientReadHandler.java
@@ -214,14 +214,14 @@ public abstract class AbstractHdfsClientReadHandler extends AbstractFileClientRe
         path = indexSegment.getPath();
         HdfsFileReader reader = indexReaderMap.get(path);
         reader.seek(indexSegment.getOffset());
-        FileBasedShuffleSegment segment = reader.readIndex();
+        FileBasedShuffleSegment segment = reader.readIndexSegment();
         while (segment != null) {
           segments.add(segment);
           size += FileBasedShuffleSegment.SEGMENT_SIZE;
           if (size >= indexSegment.getLength()) {
             break;
           } else {
-            segment = reader.readIndex();
+            segment = reader.readIndexSegment();
           }
         }
       } catch (Exception e) {
--- a/storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsFileReader.java
+++ b/storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsFileReader.java
@@ -27,6 +27,7 @@ import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.LinkedList;
 import java.util.List;
+import org.apache.commons.io.IOUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FileSystem;
@@ -76,11 +77,20 @@ public class HdfsFileReader implements ShuffleReader, Closeable {
     fsDataInputStream.seek(offset);
   }
 
+  public byte[] readIndex() {
+    try {
+      return IOUtils.toByteArray(fsDataInputStream);
+    } catch (IOException e) {
+      LOG.error("Fail to read all data from {}", path, e);
+      return new byte[0];
+    }
+  }
+
   public List<FileBasedShuffleSegment> readIndex(int limit) throws IOException, IllegalStateException {
     List<FileBasedShuffleSegment> ret = new LinkedList<>();
 
     for (int i = 0; i < limit; ++i) {
-      FileBasedShuffleSegment segment = readIndex();
+      FileBasedShuffleSegment segment = readIndexSegment();
       if (segment == null) {
         break;
       }
@@ -90,7 +100,7 @@ public class HdfsFileReader implements ShuffleReader, Closeable {
     return ret;
   }
 
-  public FileBasedShuffleSegment readIndex() throws IOException, IllegalStateException {
+  public FileBasedShuffleSegment readIndexSegment() throws IOException, IllegalStateException {
     long offset;
     long pos = fsDataInputStream.getPos();
     try {
--- a/storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsShuffleWriteHandler.java
+++ b/storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsShuffleWriteHandler.java
@@ -18,6 +18,7 @@
 
 package com.tencent.rss.storage.handler.impl;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.tencent.rss.common.ShufflePartitionedBlock;
 import com.tencent.rss.storage.common.FileBasedShuffleSegment;
 import com.tencent.rss.storage.handler.api.ShuffleWriteHandler;
@@ -131,5 +132,10 @@ public class HdfsShuffleWriteHandler implements ShuffleWriteHandler {
     Path path = new Path(basePath, fileName);
     HdfsFileWriter writer = new HdfsFileWriter(path, hadoopConf);
     return writer;
+  }
+
+  @VisibleForTesting
+  public void setFailTimes(int failTimes) {
+    this.failTimes = failTimes;
   }
 }
--- a/storage/src/main/java/com/tencent/rss/storage/handler/impl/LocalFileReader.java
+++ b/storage/src/main/java/com/tencent/rss/storage/handler/impl/LocalFileReader.java
@@ -26,6 +26,7 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.util.LinkedList;
 import java.util.List;
+import org.apache.commons.io.IOUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -52,11 +53,20 @@ public class LocalFileReader implements ShuffleReader, Closeable {
     return new byte[0];
   }
 
+  public byte[] readIndex() {
+    try {
+      return IOUtils.toByteArray(dataInputStream);
+    } catch (IOException e) {
+      LOG.error("Fail to read all data from {}", path, e);
+      return new byte[0];
+    }
+  }
+
   public List<FileBasedShuffleSegment> readIndex(int limit) throws IOException, IllegalStateException {
     List<FileBasedShuffleSegment> ret = new LinkedList<>();
 
     for (int i = 0; i < limit; ++i) {
-      FileBasedShuffleSegment segment = readIndex();
+      FileBasedShuffleSegment segment = readIndexSegment();
       if (segment == null) {
         break;
       }
@@ -66,7 +76,7 @@ public class LocalFileReader implements ShuffleReader, Closeable {
     return ret;
   }
 
-  public FileBasedShuffleSegment readIndex() throws IOException, IllegalStateException {
+  public FileBasedShuffleSegment readIndexSegment() throws IOException, IllegalStateException {
     if (dataInputStream.available() <= 0) {
       return null;
     }
--- a/storage/src/main/java/com/tencent/rss/storage/handler/impl/MultiStorageHdfsClientReadHandler.java
+++ b/storage/src/main/java/com/tencent/rss/storage/handler/impl/MultiStorageHdfsClientReadHandler.java
@@ -177,7 +177,7 @@ public class MultiStorageHdfsClientReadHandler extends AbstractHdfsClientReadHan
         path = indexSegment.getPath();
         HdfsFileReader reader = indexReaderMap.get(path);
         reader.seek(indexSegment.getOffset());
-        FileBasedShuffleSegment segment = reader.readIndex();
+        FileBasedShuffleSegment segment = reader.readIndexSegment();
         while (segment != null) {
           segment.setOffset(segment.getOffset() + indexSegment.getLastPos());
           segments.add(segment);
@@ -185,7 +185,7 @@ public class MultiStorageHdfsClientReadHandler extends AbstractHdfsClientReadHan
           if (size >= indexSegment.getLength()) {
             break;
           } else {
-            segment = reader.readIndex();
+            segment = reader.readIndexSegment();
           }
         }
       } catch (Exception e) {
--- a/storage/src/test/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandlerTest.java
+++ b/storage/src/test/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandlerTest.java
@@ -1,0 +1,154 @@
+/*
+ * Tencent is pleased to support the open source community by making
+ * Firestorm-Spark remote shuffle server available.
+ *
+ * Copyright (C) 2021 THL A29 Limited, a Tencent company.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+ * this file except in compliance with the License. You may obtain a copy of the
+ * License at
+ *
+ * https://opensource.org/licenses/Apache-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package com.tencent.rss.storage.handler.impl;
+
+import static junit.framework.Assert.assertEquals;
+import static junit.framework.Assert.assertNull;
+import static junit.framework.Assert.fail;
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertFalse;
+
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
+import com.tencent.rss.common.BufferSegment;
+import com.tencent.rss.common.ShuffleDataResult;
+import com.tencent.rss.common.ShufflePartitionedBlock;
+import com.tencent.rss.common.util.ChecksumUtils;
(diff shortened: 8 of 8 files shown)


common/src/main/java/com/tencent/rss/common/ShuffleDataSegment.java
```diff
@@ -18,23 +18,35 @@
 
 package com.tencent.rss.common;
 
+import static com.tencent.rss.common.util.Constants.DEFAULT_PATH_PREFIX;
```
jerqi, line 21: consistent with our code style?

common/src/main/java/com/tencent/rss/common/ShuffleIndexDataItem.java
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package com.tencent.rss.common;
+
+import static com.tencent.rss.common.util.Constants.DEFAULT_PATH_PREFIX;
```
jerqi, line 21: ditto

common/src/main/java/com/tencent/rss/common/ShuffleIndexResult.java
```diff
@@ -18,22 +18,47 @@
...
   }
 
-  public ShuffleIndexResult(byte[] bytes) {
-    this.indexData = bytes;
+  public ShuffleIndexResult(List<ShuffleIndexDataItem> indexDataItems) {
+    this.shuffleIndexDataItems = Lists.newArrayList();
+    for (ShuffleIndexDataItem item : indexDataItems) {
+      if (!item.isEmpty()) {
+        shuffleIndexDataItems.add(item);
+      }
```
jerqi, line 44: need we log warn some message when item is empty.

common/src/main/java/com/tencent/rss/common/ShuffleIndexDataItem.java
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * ShuffleIndexDataItem is the index data read from server side in local mode or hdfs side in hdfs mode.
+ * It is guaranteed that only one index file at most in the server side in local mode, in which the
+ * dataPathPrefix filed is useless and set to default value. However there may be multiple index file in hdfs mode,
+ * and the dataPathPrefix the dataPathPrefix field should be a valid data path prefix can be used to get the
+ * data file indexed by this index item.
+ */
+public class ShuffleIndexDataItem {
```
jerqi, line 30: may ShuffleIndexItem be better? Because index and data is two concepts in our system, we shouldn't mix them.

jerqi: I feel that ShuffleIndexDataItem should be used in hdfs mode only.  In local mode, prefix path is weird.

common/src/main/java/com/tencent/rss/common/ShuffleDataSegment.java
duanmeng: sure

common/src/main/java/com/tencent/rss/common/ShuffleIndexDataItem.java
duanmeng: what dose ditto mean?

common/src/main/java/com/tencent/rss/common/ShuffleIndexResult.java
duanmeng: OK

common/src/main/java/com/tencent/rss/common/ShuffleIndexDataItem.java
duanmeng: Agree to use ShuffleIndexItem instead. But path prefix is a common concept in both local and hdfs, this field can be empty in local is simple because we force only one data/index file to exist.

common/src/main/java/com/tencent/rss/common/ShuffleIndexDataItem.java
jerqi: The same with above.

common/src/main/java/com/tencent/rss/common/ShuffleIndexDataItem.java
jerqi: But path inl local mode is  invisible to client.

colinmjj:
There are different ways to have data replica with Local and HDFS.
With Local mode, data should be sent to different server, the read process should be successfully even one shuffle server failed.
With HDFS mode, HDFS should guarantee that data won't miss.
So, why this PR is necessary?

jerqi:
> There are different ways to have data replica with Local and HDFS. With Local mode, data should be sent to different server, the read process should be successfully even one shuffle server failed. With HDFS mode, HDFS should guarantee that data won't miss. So, why this PR is necessary?

If we read index first, we can skip necassary blockID  like produced in aqe.

storage/src/main/java/com/tencent/rss/storage/api/ShuffleReader.java
```diff
@@ -26,5 +26,8 @@
 
   byte[] readData(long offset, int length);
 
+  byte[] readAllData();
```
colinmjj, line 29: This api is unnecessary, readIndex(int size) is ok here,  return type can be changed if necessary.

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
```diff
@@ -51,7 +61,58 @@ public HdfsClientReadHandler(
...
+    ShuffleDataResult shuffleDataResult = null;
+    HdfsFileReader reader = dataReaderMap.get(shuffleDataSegment.getPathPrefix());
+    if (reader == null) {
+      LOG.error("Invalid segment {}", shuffleDataSegment);
+      return shuffleDataResult;
+    }
+
+    byte[] data = reader.readData(shuffleDataSegment.getOffset(), shuffleDataSegment.getLength());
+    return new ShuffleDataResult(data, shuffleDataSegment.getBufferSegments());
+
```
colinmjj, line 98: remove empty line

common/src/main/java/com/tencent/rss/common/ShuffleIndexDataItem.java
colinmjj: here is the problem, if only one data/index file is possible?
The current implementation is the work around if can't append data to hdfs, eg, lease problem
It may be time to improve it first

duanmeng:
> There are different ways to have data replica with Local and HDFS. With Local mode, data should be sent to different server, the read process should be successfully even one shuffle server failed. With HDFS mode, HDFS should guarantee that data won't miss. So, why this PR is necessary?

We need to read index first to make it possible to read data on demand and current read predure is not efficient for it have to read index file twice.

storage/src/main/java/com/tencent/rss/storage/api/ShuffleReader.java
duanmeng: This is a readAll api or we can read all data in readData if the length is -1

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
duanmeng: OK

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
```diff
@@ -51,7 +63,83 @@ public HdfsClientReadHandler(
...
+    if (!cursor.hasNext()) {
+      return null;
+    }
+
+    Entry<String, HdfsFileReader> entry = cursor.next();
+    curPath = entry.getKey();
+    long start = System.currentTimeMillis();
+    HdfsFileReader reader = entry.getValue();
+    if (reader == null) {
+      LOG.error("Reader of {} is invalid!", curPath);
```
jerqi, line 79: return null?

duanmeng: OK

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
```diff
@@ -51,7 +66,91 @@ public HdfsClientReadHandler(
...
+        .iterator();
+  }
+
+  public ShuffleIndexResult readShuffleIndex() {
+    if (!cursor.hasNext()) {
+      return null;
+    }
+
+    Entry<String, HdfsFileReader> entry = cursor.next();
+    curPath = entry.getKey();
```
jerqi, line 83: currentPath?cursorPath?

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
```diff
@@ -51,7 +66,91 @@ public HdfsClientReadHandler(
...
+
+  public ShuffleDataResult readShuffleData(ShuffleDataSegment shuffleDataSegment) {
+    HdfsFileReader reader = dataReaderMap.get(curPath);
+    if (reader == null) {
+      LOG.error("Invalid segment {}", shuffleDataSegment);
+      return null;
+    }
+
+    int expectedLength = shuffleDataSegment.getLength();
+    if (expectedLength <= 0) {
```
jerqi, line 111: if (shuffleDataSegment.getLength() < 0)

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
```diff
@@ -51,7 +66,91 @@ public HdfsClientReadHandler(
...
+      LOG.error("Invalid segment {}", shuffleDataSegment);
+      return null;
+    }
+
+    byte[] data = reader.readData(shuffleDataSegment.getOffset(), shuffleDataSegment.getLength());
+    if (data.length != expectedLength) {
```
jerqi, line 117: if (data.length != shuffleDataSegment.getLength())

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
duanmeng: OK

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
duanmeng: currentPath

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
```diff
@@ -28,6 +39,10 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(HdfsClientReadHandler.class);
 
+  private Iterator<Entry<String, HdfsFileReader>> cursor;
```
jerqi, line 42: readerIterator?

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
duanmeng: expectedLength is useful in the following code

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
```diff
@@ -51,7 +66,96 @@ public HdfsClientReadHandler(
         ShuffleStorageUtils.getShuffleDataPathWithRange(appId,
             shuffleId, partitionId, partitionNumPerRange, partitionNum));
     init(fullShufflePath);
-    readAllIndexSegments();
+    cursor = indexReaderMap
+        .entrySet()
+        .stream()
+        .sorted(Comparator.comparing(Entry::getKey))
```
jerqi, line 72: why we need sort?

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
duanmeng: unified the use of expectedLength to avoid multiple call of getLength()

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
```diff
@@ -51,7 +66,96 @@ public HdfsClientReadHandler(
...
+      return null;
+    }
+
+    Entry<String, HdfsFileReader> entry = cursor.next();
+    cursorPath = entry.getKey();
+    long start = System.currentTimeMillis();
+    HdfsFileReader reader = entry.getValue();
+    if (reader == null) {
+      LOG.error("Reader of {} is invalid!", cursorPath);
+      return null;
```
jerqi, line 88: Should we return null or empty ShuffleIndexResult?  Can we unify these?

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
```diff
@@ -51,7 +66,96 @@ public HdfsClientReadHandler(
...
+        .entrySet()
+        .stream()
+        .sorted(Comparator.comparing(Entry::getKey))
+        .collect(Collectors.toList())
+        .iterator();
+  }
+
+  public ShuffleIndexResult readShuffleIndex() {
+    if (!cursor.hasNext()) {
+      return null;
```
jerqi, line 79: 2

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
```diff
@@ -51,7 +66,96 @@ public HdfsClientReadHandler(
...
+    try {
+      byte[] indexData = reader.readAllData();
+      LOG.info("Read index files for shuffleId[{}], partitionId[{}] for {} ms",
+          shuffleId, partitionId, System.currentTimeMillis() - start);
+      return new ShuffleIndexResult(indexData);
+    } catch (Exception e) {
+      LOG.info("Fail to read index files for shuffleId[{}], partitionId[{}]", shuffleId, partitionId);
+    }
+
+    return new ShuffleIndexResult();
```
jerqi, line 100: 3

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
```diff
@@ -51,7 +66,96 @@ public HdfsClientReadHandler(
...
-    readAllIndexSegments();
+    cursor = indexReaderMap
+        .entrySet()
+        .stream()
+        .sorted(Comparator.comparing(Entry::getKey))
+        .collect(Collectors.toList())
+        .iterator();
+  }
+
+  public ShuffleIndexResult readShuffleIndex() {
```
jerqi, line 77: Don't we read all indexFile like local?

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
duanmeng: Read file by time sequnce.

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
duanmeng: Yes, we read a data of a index at once as local mode. But we read index files one by one for there may be multiple index/data files in hdfs mode.

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
jerqi: What's advantage?

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
duanmeng: Ok, this would make readShuffleData remove the null conditiion check.

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
duanmeng: 1. read earlier write early
2. read the group of files written by the same server

jerqi: `ls` will return dictionary-order file list, is it enough?

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
jerqi: Can we read all index file?

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
duanmeng: It may not be appropriate to rely on hadoop list status api's behavior nor the map's iterator sequnce (and what's more it's hash map).

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
duanmeng: We must maintain the mapping relation between path and the shuffle index segment if we read all index file first like the first versoin of this pr. And we don't need to read the replica index if all expected blocks have been read.

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
duanmeng: ok

common/src/main/java/com/tencent/rss/common/util/Constants.java
```diff
@@ -32,4 +32,5 @@
   public static long MAX_SEQUENCE_NO = (1 << Constants.ATOMIC_INT_MAX_LENGTH) - 1;
   public static long MAX_PARTITION_ID = (1 << Constants.PARTITION_ID_MAX_LENGTH) - 1;
   public static long MAX_TASK_ATTEMPT_ID = (1 << Constants.TASK_ATTEMPT_ID_MAX_LENGTH) - 1;
+  public static final String DEFAULT_PATH_PREFIX = "/dev/null/prefix";
```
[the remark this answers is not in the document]
duanmeng: removed

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
```diff
@@ -51,7 +66,96 @@ public HdfsClientReadHandler(
...
+    return readShuffleData(shuffleDataSegments.get(segmentIndex));
   }
 
+  @VisibleForTesting
+  public List<ShuffleDataSegment> getShuffleDataSegments() {
+    return shuffleDataSegments;
+  }
+
+  @VisibleForTesting
+  public Iterator<Entry<String, HdfsFileReader>> getCursor() {
```
jerqi, line 158: getReaderIterator()

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
jerqi: OK

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
jerqi: OK

storage/src/main/java/com/tencent/rss/storage/api/ShuffleReader.java
jerqi: I also think we don't need this interface.

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
```diff
@@ -51,7 +66,96 @@ public HdfsClientReadHandler(
...
+    currentPath = entry.getKey();
+    long start = System.currentTimeMillis();
+    HdfsFileReader reader = entry.getValue();
+    if (reader == null) {
+      LOG.error("Reader of {} is invalid!", currentPath);
+      return new ShuffleIndexResult();
+    }
+
+    try {
+      byte[] indexData = reader.readAllData();
```
jerqi, line 92: why do we use readAllData instead of readIndex?

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
jerqi: Can we create a class named like 'HdfsShuffleIndexResult' to record the mapping relation.

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
```diff
@@ -28,6 +39,10 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(HdfsClientReadHandler.class);
 
+  private final Iterator<Entry<String, HdfsFileReader>> readerIterator;
+  private final List<ShuffleDataSegment> shuffleDataSegments = Lists.newArrayList();
+  private String currentPath = "";
```
jerqi, line 44: currentPath is a little weird. I feel we should have a new class to record the mapping relation, and remove currentPath.

storage/src/main/java/com/tencent/rss/storage/api/ShuffleReader.java
colinmjj: readAll should be renamed as readIndex?

storage/src/main/java/com/tencent/rss/storage/api/ShuffleReader.java
```diff
@@ -26,5 +26,7 @@
...
+
   List<FileBasedShuffleSegment> readIndex(int limit) throws IOException, IllegalStateException;
```
colinmjj, line 31: does this api needed?

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
duanmeng: I used to consider this way, but it's kind of overdesign since we have maintain the path to file relation inside the handler. And we would have to add another class HdfsShuffleDataSegment if the class is used.

storage/src/main/java/com/tencent/rss/storage/api/ShuffleReader.java
duanmeng: Agree, I think we need a readIndexRawData api here, readAllData may break the sematic consistence of of this interface.

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
duanmeng: The same as above

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
duanmeng: OK

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
duanmeng: Will add a iterator wrapper class with getKey, next, hasNext.

storage/src/main/java/com/tencent/rss/storage/api/ShuffleReader.java
jerqi: I think `readIndex` may express the behavior more correctly that we want. Our `FileReader` present `ShuffleFIleReader`, not a common file reader. So the `FileReader` can have the `readIndex` interface.

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
jerqi: Why is  this overdesign?

duanmeng: Both local file and hdfs file has a path. I prefer to have a ShuffleIndexResult with path field than having a HdfsShuffleIndexResult just to add  a common filed, which is implemented in last version.

storage/src/main/java/com/tencent/rss/storage/api/ShuffleReader.java
duanmeng: OK, add a readIndex api that return raw data of the index file.

storage/src/main/java/com/tencent/rss/storage/api/ShuffleReader.java
duanmeng: It's used in AbstractHdfsClientReadHandler and MultiStorageHdfsClientReadHandler, I will not make any change in this pr. Will tidy this interface in the following pr to refactor reading in multi mode.

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
jerqi: We don't care the path for local file, do it?

duanmeng: This is because the relation maintained by the server handler and hdfs file path is maintained by the hdfs read handler now and no need to add a HdfsShuffleIndexResult and HdfsShuffleDataSegment to add path.

jerqi: Could you paste the latest  code?

duanmeng: No, I overwrite the last version by current solution since you commented path is appropriate and I agreed with that.

storage/src/main/java/com/tencent/rss/storage/api/ShuffleReader.java
```diff
@@ -26,5 +26,7 @@
 
   byte[] readData(long offset, int length);
 
+  byte[] readIndexRawData();
```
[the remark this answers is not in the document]
duanmeng: readIndex is the old way to read index data, parsed to index segment before return. so we need a api.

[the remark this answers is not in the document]
jerqi: readIndexRawData -> readIndex? Because I don't care the index data whether is raw or parsed.

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
jerqi: Why don't add HdfsShuffleIndexResult? Although we just add a single field, we still have a property abstraction.

storage/src/main/java/com/tencent/rss/storage/api/ShuffleReader.java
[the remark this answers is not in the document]
duanmeng: No you must care, `byte[] readIndex()` have the same erasure as `FileBasedShuffleSegment readIndex()` in all shuffle readers.

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
duanmeng: In this version we don't need it anymore.

storage/src/main/java/com/tencent/rss/storage/api/ShuffleReader.java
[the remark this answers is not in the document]
jerqi: Our method is `List<FileBasedShuffleSegment> readIndex(int limit)` not `byte[] readIndex()`. They have different erasure.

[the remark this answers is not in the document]
duanmeng: Will replace readIndex in all shuffle readers. And change readIndexRawData to readIndex to make this interface more consistence.

storage/src/main/java/com/tencent/rss/storage/handler/impl/HdfsClientReadHandler.java
jerqi: Actually, I wanna to read all index file, so the mapping relation is necessary.

duanmeng: This mapping always needed no matter your read pattern or mode. It's maintained in the handler. And I make it clear at the beginning that I read the whole index file at once and read the file one by one.

jerqi: I hope to read all files? WDYT@colinma

colinmjj:
@duanmeng @jerqi according to offline discussion, this PR is only target to improve how to read data with HDFS. The code should be refactor with another PR for better architecture of ReadHandler

colinmjj (approved):
LGTM

PR #24, opened 2021-11-24
[Feature] Add the metrics for the number of total partitions

jerqi:
Add a metrics that represent the number of total partitions in a server.

The total number of  partitions is useful for us. It can reflect the load of the shuffle server. If a shuffle server has too many partitions, the shuffle server will cause too much random io, the shuffle server will have poor performance.

No

Manual test

### DIFF
--- a/server/src/main/java/com/tencent/rss/server/ShuffleBufferManager.java
+++ b/server/src/main/java/com/tencent/rss/server/ShuffleBufferManager.java
@@ -157,10 +157,10 @@ public class ShuffleBufferManager {
     long size = 0;
     for (RangeMap<Integer, ShuffleBuffer> rangeMap : shuffleIdToBuffers.values()) {
       if (rangeMap != null) {
-        ShuffleServerMetrics.gaugeTotalPartitionNum.dec();
         Collection<ShuffleBuffer> buffers = rangeMap.asMapOfRanges().values();
         if (buffers != null) {
           for (ShuffleBuffer buffer : buffers) {
+            ShuffleServerMetrics.gaugeTotalPartitionNum.dec();
             size += buffer.getSize();
           }
         }
--- a/server/src/main/java/com/tencent/rss/server/ShuffleBufferManager.java
+++ b/server/src/main/java/com/tencent/rss/server/ShuffleBufferManager.java
@@ -66,6 +66,7 @@ public class ShuffleBufferManager {
     shuffleIdToBuffers.putIfAbsent(shuffleId, TreeRangeMap.create());
     RangeMap<Integer, ShuffleBuffer> bufferRangeMap = shuffleIdToBuffers.get(shuffleId);
     if (bufferRangeMap.get(startPartition) == null) {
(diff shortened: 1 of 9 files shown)


colinmjj:
It should be a common metrics, not for multiple storage only.

jerqi:
> It should be a common metrics, not for multiple storage only.

Only multiStorageManager has the number of partitions.

colinmjj:
> > It should be a common metrics, not for multiple storage only.
> 
> Only multiStorageManager has the number of partitions.

I think the metrics is about how many ShuffleBuffer in ShuffleServer

jerqi:
> > > It should be a common metrics, not for multiple storage only.
> > 
> > 
> > Only multiStorageManager has the number of partitions.
> 
> I think the metrics is about how many ShuffleBuffer in ShuffleServer

ShuffleBuffer's collection is RangeMap. RangeMap don't have method `size`

colinmjj:
> > > > It should be a common metrics, not for multiple storage only.
> > > 
> > > 
> > > Only multiStorageManager has the number of partitions.
> > 
> > 
> > I think the metrics is about how many ShuffleBuffer in ShuffleServer
> 
> ShuffleBuffer's collection is RangeMap. RangeMap don't have method `size`

For this PR, it should target on how to collect the size of current buffer which will be written to single file.
It's a common metrics, if there has some problem in current implementation, we should do the optimization, not just ignore it.

jerqi:
> > > > > It should be a common metrics, not for multiple storage only.
> > > > 
> > > > 
> > > > Only multiStorageManager has the number of partitions.
> > > 
> > > 
> > > I think the metrics is about how many ShuffleBuffer in ShuffleServer
> > 
> > 
> > ShuffleBuffer's collection is RangeMap. RangeMap don't have method `size`
> 
> For this PR, it should target on how to collect the size of current buffer which will be written to single file. It's a common metrics, if there has some problem in current implementation, we should do the optimization, not just ignore it.

OK, fixed.

server/src/main/java/com/tencent/rss/server/ShuffleBufferManager.java
```diff
@@ -156,6 +157,7 @@ public void removeBuffer(String appId) {
     long size = 0;
     for (RangeMap<Integer, ShuffleBuffer> rangeMap : shuffleIdToBuffers.values()) {
       if (rangeMap != null) {
+        ShuffleServerMetrics.gaugeTotalPartitionNum.dec();
         Collection<ShuffleBuffer> buffers = rangeMap.asMapOfRanges().values();
         if (buffers != null) {
           for (ShuffleBuffer buffer : buffers) {
```
colinmjj, line 163: dec() should be called in this loop?

jerqi: Yes.

jerqi: My mistake. I will Fix this.

colinmjj (approved):
LGTM

PR #27, opened 2021-11-30
[Improvement] Modify some configuration's names and default values

jerqi:
Modify some configuration's names and default values.

Unify configuration option style, and update proper default value for configuration.

No

GA passed


server/src/main/java/com/tencent/rss/server/ShuffleServerConf.java
```diff
@@ -123,7 +123,7 @@
   public static final ConfigOption<Long> SERVER_PRE_ALLOCATION_EXPIRED = ConfigOptions
       .key("rss.server.preAllocation.expired")
       .longType()
-      .defaultValue(10 * 1000L)
+      .defaultValue(120 * 1000L)
```
colinmjj, line 126: 120 is too long for preAllocation expired

jerqi: In sop and online ,  we choose 120s. What's your suggestion?

colinmjj: when client send data to shuffle server, request memory first, then send data if there has enough memory. Shuffle server will keep the preAllocation memory until preAllocation.expired. If application is killed, there will be a lot of preAllocation memory in shuffle server.
preAllocation.expired is about interval of 2 rpc calls. If 10s is too short, it means there has some problem in rpc communication, maybe overload, etc.  These problem should be resolved first, preAllocation.expired is just a workaround for these problem.
I suggest to set 20s as default value at most

common/src/main/java/com/tencent/rss/common/config/RssBaseConf.java
```diff
@@ -110,7 +110,7 @@
   public static final ConfigOption<Integer> RPC_MESSAGE_MAX_SIZE = ConfigOptions
       .key("rss.rpc.message.max.size")
       .intType()
-      .defaultValue(Integer.MAX_VALUE)
+      .defaultValue(1073741824)
```
duanmeng, line 113: 1073741824 -> 1024 * 1024  * 1024 to make it readable.

server/src/main/java/com/tencent/rss/server/ShuffleServerConf.java
jerqi: OK

common/src/main/java/com/tencent/rss/common/config/RssBaseConf.java
jerqi: OK

jerqi:
@duanmeng @colinmjj all comments are addressed.

jerqi:
@duanmeng @colinmjj gently ping.

colinmjj (approved):
LGTM

PR #28, opened 2021-12-02
Improve lock strategy to reduce competition when report shuffle result

colinmjj:
Improve lock strategy to reduce competition when report shuffle result

Current implementation will cause much latency for lock competition
According to the simulator test, the improvement can get 4X performance when report shuffle result

NO

With current UT

### DIFF
--- a/server/src/main/java/com/tencent/rss/server/ShuffleTaskManager.java
+++ b/server/src/main/java/com/tencent/rss/server/ShuffleTaskManager.java
@@ -35,6 +35,11 @@ import com.tencent.rss.common.util.RssUtils;
 import com.tencent.rss.storage.factory.ShuffleHandlerFactory;
 import com.tencent.rss.storage.handler.api.ServerReadHandler;
 import com.tencent.rss.storage.request.CreateShuffleReadHandlerRequest;
+import org.roaringbitmap.longlong.LongIterator;
+import org.roaringbitmap.longlong.Roaring64NavigableMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


server/src/main/java/com/tencent/rss/server/ShuffleTaskManager.java
```diff
@@ -207,12 +208,32 @@ public void addFinishedBlockIds(
...
+
+    // This method will be called from many threads, it's important to keep lock in a reasonable way
+    // Here is the strategy:
+    // 1. get lock of bitmap
+    // 2. update the bitmap
+    // 3. do 1 & 2 again for another bitmap
+    // Todo: High latency with current implementation for the lock competition
+    List<List<long[]>> entryList = Lists.newArrayList();
+    for (int i = 0; i < bitmapNum; i++) {
+      entryList.add(i, Lists.newArrayList());
```
jerqi, line 220: We should a meaningful name, not entryList

server/src/main/java/com/tencent/rss/server/ShuffleTaskManager.java
```diff
@@ -207,12 +208,32 @@ public void addFinishedBlockIds(
...
     for (Map.Entry<Integer, long[]> entry : partitionToBlockIds.entrySet()) {
       Integer partitionId = entry.getKey();
-      Roaring64NavigableMap bitmap = blockIds[partitionId % bitmapNum];
+      entryList.get(partitionId % bitmapNum).add(entry.getValue());
+    }
+
+    for (int i = 0; i < bitmapNum; i++) {
+      Roaring64NavigableMap bitmap = blockIds[i];
+      List<long[]> blockIdList = entryList.get(i);
       synchronized (bitmap) {
```
jerqi, line 232: To be honesty, I don't ensure that these will solve the problem. We should some mircro benchmark at least.

colinmjj: Yes，according to the simulator test, the improvement can get 4X performance

jerqi: Can you describe some details about the simulator test?

jerqi:
I test this situation, this can't improve the performance.I will close this pr.

PR #30, opened 2021-12-08
[Feature]Add local storage metrics

duanmeng:
Add local storage metric to track total data size of local dir.

We need this metric to help analyze the workload.

No

UT and integration test.

### DIFF
--- a/server/src/main/java/com/tencent/rss/server/LocalStorageChecker.java
+++ b/server/src/main/java/com/tencent/rss/server/LocalStorageChecker.java
@@ -61,11 +61,14 @@ public class LocalStorageChecker extends Checker {
   @Override
   public boolean checkIsHealthy() {
     int num = 0;
+    long totalLocalDataSize = 0;
     for (StorageInfo storageInfo : storageInfos) {
       if (storageInfo.checkIsHealthy()) {
         num++;
       }
+      totalLocalDataSize += storageInfo.getUsedSize();
     }
+    ShuffleServerMetrics.gaugeTotalLocalDataSize.set(totalLocalDataSize);
 
     if (storageInfos.isEmpty()) {
       if (isHealthy) {
@@ -107,6 +110,7 @@ public class LocalStorageChecker extends Checker {
 
     private final File storageDir;
     private boolean isHealthy;
+    private long usedSize;
 
     StorageInfo(String path) {
       this.storageDir = new File(path);
@@ -118,7 +122,10 @@ public class LocalStorageChecker extends Checker {
         this.isHealthy = false;
         return false;
       }
-      double usagePercent = getUsedSpace(storageDir) * 100.0 / getTotalSpace(storageDir);
+      // We assume that this partition is only used by us so the used space is the local shuffle data size.
+      // If not we should walk through all the files and get sum of all file length.
+      usedSize = getUsedSpace(storageDir);
+      double usagePercent = usedSize * 100.0 / getTotalSpace(storageDir);
       if (isHealthy) {
         if (Double.compare(usagePercent, diskMaxUsagePercentage) >= 0) {
           isHealthy = false;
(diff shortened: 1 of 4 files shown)


server/src/main/java/com/tencent/rss/server/LocalStorageChecker.java
```diff
@@ -61,11 +61,14 @@ public LocalStorageChecker(ShuffleServerConf conf) {
   @Override
   public boolean checkIsHealthy() {
     int num = 0;
+    long totalLocalDataSize = 0;
     for (StorageInfo storageInfo : storageInfos) {
       if (storageInfo.checkIsHealthy()) {
         num++;
       }
+      totalLocalDataSize += getUsedSpace(storageInfo.storageDir);
```
colinmjj, line 69: getUsedSpace is called in storageInfo.checkIsHealthy(), reuse it if it's a heavy call which impact performance

duanmeng: OK, add a filed named usedSize in the StorageInfo？

server/src/main/java/com/tencent/rss/server/ShuffleServerMetrics.java
```diff
@@ -52,6 +52,8 @@
   private static final String TOTAL_UPLOAD_SIZE = "total_upload_size";
   private static final String TOTAL_UPLOAD_TIME_S = "total_upload_time_s";
   private static final String TOTAL_DROPPED_EVENT_NUM = "total_dropped_event_num";
+  private static final String TOTAL_LOCAL_DATA_SIZE = "total_local_data_size";
+  private static final String TOTAL_REMOTE_DATA_SIZE = "total_remote_data_size";
```
jerqi, line 56: remove TOTAL_REMOTE_DATA_SIZE?

duanmeng: OK

server/src/main/java/com/tencent/rss/server/LocalStorageChecker.java
```diff
@@ -107,6 +110,7 @@ long getUsedSpace(File file) {
 
     private final File storageDir;
     private boolean isHealthy;
+    private long usedSize;
```
colinmjj, line 113: it can be named as usedSpace, and be set in getUsedSpace, no extra getXXX should be added

server/src/main/java/com/tencent/rss/server/LocalStorageChecker.java
```diff
@@ -118,7 +122,10 @@ boolean checkIsHealthy() {
         this.isHealthy = false;
         return false;
       }
-      double usagePercent = getUsedSpace(storageDir) * 100.0 / getTotalSpace(storageDir);
+      // We assume that this partition is only used by us so the used space is the local shuffle data size.
```
colinmjj, line 125: Calculate the used space, the comment is not necessary for such simple case.

server/src/main/java/com/tencent/rss/server/LocalStorageChecker.java
duanmeng: `getUsedSpace` would calculate the used space of partition, we can't resue the result if rename and use getUseSpace.

server/src/main/java/com/tencent/rss/server/LocalStorageChecker.java
duanmeng: `getUsedSpace` and `getTotalSpace` actually get the partition usage info, which is not the actual data size in the storage base path. So I add a comment here to make an explanation about it.

colinmjj: comment: Get size of shuffle data, the value maybe incorrect if the folder is used by another propose.

duanmeng: ok

jerqi:
@duanmeng @duanmeng Does anything need to change？

duanmeng:
The `LocalStorageChecker` actually uses the partition size, it may not be precise, do not develop this feature for now and close this pr.

PR #31, opened 2021-12-09
[Improvement] Avoid speculative tasks causing writing errors after shuffle is uploaded and deleted

jerqi:
When event write data in LocalFileWriteHanlder, if base directorys don't exist,  the method will return directly, not throw FileNotFoundException.

In multiStorage mode, shuffle will be started upload after shuffle is read. When shuffle has been uploaded totally,  shuffle directory will be deleted. Speculative tasks may also write its data to shuffle server.So these data will cause FileNotFoundExceptions, these exceptions will degrade performance.

No.

Add newly UTS.

### DIFF
--- a/storage/src/main/java/com/tencent/rss/storage/handler/impl/LocalFileWriteHandler.java
+++ b/storage/src/main/java/com/tencent/rss/storage/handler/impl/LocalFileWriteHandler.java
@@ -88,17 +88,18 @@ public class LocalFileWriteHandler implements ShuffleWriteHandler {
   @Override
   public synchronized void write(
       List<ShufflePartitionedBlock> shuffleBlocks) throws IOException, IllegalStateException {
-    long accessTime = System.currentTimeMillis();
-    String dataFileName = ShuffleStorageUtils.generateDataFileName(fileNamePrefix);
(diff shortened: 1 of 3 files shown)


storage/src/main/java/com/tencent/rss/storage/handler/impl/LocalFileWriteHandler.java
```diff
@@ -92,6 +92,14 @@ public synchronized void write(
     String dataFileName = ShuffleStorageUtils.generateDataFileName(fileNamePrefix);
     String indexFileName = ShuffleStorageUtils.generateIndexFileName(fileNamePrefix);
 
+    // if shuffle has been uploaded totally, the directory of shuffle will be deleted
+    // the event should be dropped.
```
duanmeng, line 96: // if shuffle has been uploaded totally... -> // Ignore this write, if the shuffle directory is deleted after being uploaded in multi mode or after its app heartbeat times out.

And it may be better to put this check before the `System.currentTimeMillis()`

jerqi: OK

jerqi:
@duanmeng Does anything need to change？

duanmeng (approved):
LGTM

PR #33, opened 2021-12-13
[Improvement]Remove useless code in test base

duanmeng:
Remove the useless code that creat an extra coordinator server

These codes are useless and kind of hardcode.

No

ut and integration test

### DIFF
--- a/integration-test/common/src/test/java/com/tencent/rss/test/IntegrationTestBase.java
+++ b/integration-test/common/src/test/java/com/tencent/rss/test/IntegrationTestBase.java
@@ -96,9 +96,6 @@ abstract public class IntegrationTestBase extends HdfsTestBase {
 
   protected static void createCoordinatorServer(CoordinatorConf coordinatorConf) throws Exception {
     coordinators.add(new CoordinatorServer(coordinatorConf));
-    coordinatorConf.setInteger("rss.rpc.server.port", COORDINATOR_PORT_2);
-    coordinatorConf.setInteger("rss.jetty.http.port", JETTY_PORT_2);
-    coordinators.add(new CoordinatorServer(coordinatorConf));
   }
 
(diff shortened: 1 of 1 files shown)


colinmjj (approved):
LGTM

integration-test/common/src/test/java/com/tencent/rss/test/IntegrationTestBase.java
```diff
@@ -96,9 +96,6 @@ protected static ShuffleServerConf getShuffleServerConf() {
 
   protected static void createCoordinatorServer(CoordinatorConf coordinatorConf) throws Exception {
     coordinators.add(new CoordinatorServer(coordinatorConf));
-    coordinatorConf.setInteger("rss.rpc.server.port", COORDINATOR_PORT_2);
-    coordinatorConf.setInteger("rss.jetty.http.port", JETTY_PORT_2);
```
jerqi, line 100: we should need two coordinators

colinmjj: @jerqi yes, we should add case to test multiple coordinator.

jerqi: There are so many errors in ut's log that fail to send heartheat to secondary coordinator. @colinmjj @duanmeng If you don't have strong opinions on this, i will revert this patch.

colinmjj: @jerqi I think it's UT's problem, and ut should be fixed.
We shouldn't enable 2 coordinators with all situations.

jerqi: OK, Let's remove extra client.

PR #36, opened 2021-12-15
Support store shuffle data in memory

colinmjj:
This pr is target to store shuffle data in memory. New flush strategy and read strategy are involved. The benefits with this change are:
1. Huge memory is useful now
2. There has no wait for data flush after stage finish
3. Reduce flush data to storage

new RPC interface is involved

Yes, user need to update configuration for the new feature

UT add


integration-test/spark-common/src/test/java/com/tencent/rss/test/RepartitionWithMemoryRssTest.java
```diff
@@ -0,0 +1,60 @@
...
+    shuffleServerConf.setString(ShuffleServerConf.SERVER_BUFFER_CAPACITY.key(), "512mb");
+    shuffleServerConf.setString(ShuffleServerConf.SERVER_MEMORY_SHUFFLE_LOWWATERMARK.key(), "100mb");
+    shuffleServerConf.setString(ShuffleServerConf.SERVER_MEMORY_SHUFFLE_HIGHWATERMARK.key(), "400mb");
+    createShuffleServer(shuffleServerConf);
+    startServers();
+  }
+
+  @Override
+  public void updateRssStorage(SparkConf sparkConf) {
+    sparkConf.set(RssClientConfig.RSS_STORAGE_TYPE, "MEMORY,LOCALFILE");
```
jerqi, line 58: MEMORY_AND_LOCAILFILE?

colinmjj: Current style for flexible without adding new storage type, eg, "MEMORY,HDFS" means memory and hdfs, "LOCAL,HDFS" means local and hdfs

jerqi: I think multi storage should separate from single storage. MEMORY_AND_LOCALFILE is more Spark's style.And i can't find more discussion on this in Spark's pr.

colinmjj: Agree to refactor multi storage, I think it can reuse ComposedClientReadHandler, but it should be another PR

jerqi: storageType need to be split with `,`. It's a little weird.

server/src/main/java/com/tencent/rss/server/buffer/FSShuffleBufferManager.java
```diff
@@ -0,0 +1,75 @@
...
+import com.google.common.collect.RangeMap;
+import com.tencent.rss.common.ShuffleDataResult;
+import com.tencent.rss.server.ShuffleFlushManager;
+import com.tencent.rss.server.ShuffleServerConf;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Map;
+
+public class FSShuffleBufferManager extends ShuffleBufferManager {
```
duanmeng, line 32: The FS acronym may be not clear enough as a class name.

jerqi:
This patch is too big, can you split this pr to small prs?

storage/src/main/java/com/tencent/rss/storage/handler/impl/ComposedClientReadHandler.java
```diff
@@ -0,0 +1,109 @@
...
+import org.slf4j.LoggerFactory;
+
+public class ComposedClientReadHandler implements ClientReadHandler {
+
+  private static final Logger LOG = LoggerFactory.getLogger(ComposedClientReadHandler.class);
+
+  private ClientReadHandler level1ReadHandler;
+  private ClientReadHandler level2ReadHandler;
+  private ClientReadHandler level3ReadHandler;
+  private static final int LEVEL1 = 1;
```
duanmeng, line 33: Could you use enum and add some comments here, since the level with number naming convetion is hard to understand.

colinmjj:
> 

how about create a new branch for memory shuffle, and try to split the pr

storage/src/main/java/com/tencent/rss/storage/handler/impl/MemoryClientReadHandler.java
```diff
@@ -0,0 +1,98 @@
...
+import com.tencent.rss.common.BufferSegment;
+import com.tencent.rss.common.ShuffleDataResult;
+import com.tencent.rss.common.exception.RssException;
+import com.tencent.rss.common.util.Constants;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+
+public class MemoryClientReadHandler extends AbstractFileClientReadHandler {
```
duanmeng, line 34: IMHO it may not be appropriate to have a sepecific *Memory* handler here, the client read data either from server or remote fs directly and it should not know storage deatils in the server side when read data from the server.

duanmeng:
> > 
> 
> how about create a new branch for memory shuffle, and try to split the pr

Good idea, it's so huge:)

integration-test/spark-common/src/test/java/com/tencent/rss/test/SparkSQLWithMemoryRssTest.java
```diff
@@ -0,0 +1,65 @@
...
+    shuffleServerConf.set(ShuffleServerConf.SERVER_HEARTBEAT_INTERVAL, 5000L);
+    shuffleServerConf.set(ShuffleServerConf.SERVER_APP_EXPIRED_WITHOUT_HEARTBEAT, 4000L);
+    File tmpDir = Files.createTempDir();
+    tmpDir.deleteOnExit();
+    File dataDir1 = new File(tmpDir, "data1");
+    File dataDir2 = new File(tmpDir, "data2");
+    String basePath = dataDir1.getAbsolutePath() + "," + dataDir2.getAbsolutePath();
+    shuffleServerConf.set(ShuffleServerConf.RSS_STORAGE_TYPE, StorageType.LOCALFILE.name());
+    shuffleServerConf.set(ShuffleServerConf.RSS_STORAGE_BASE_PATH, basePath);
+    shuffleServerConf.set(ShuffleServerConf.SERVER_MEMORY_SHUFFLE_ENABLED, true);
```
jerqi, line 48: Why we can't unify the configuration between server and client?

proto/src/main/proto/Rss.proto
```diff
@@ -29,6 +29,7 @@ service ShuffleServer {
   rpc sendShuffleData (SendShuffleDataRequest) returns (SendShuffleDataResponse);
   rpc getShuffleIndex (GetShuffleIndexRequest) returns (GetShuffleIndexResponse);
   rpc getShuffleData (GetShuffleDataRequest) returns (GetShuffleDataResponse);
+  rpc getInMemoryShuffleData (GetInMemoryShuffleDataRequest) returns (GetInMemoryShuffleDataResponse);
```
jerqi, line 32: Why can we reuse getShuffleData and getShuffleIndex?

server/src/main/java/com/tencent/rss/server/ShuffleServerMetrics.java
```diff
@@ -53,33 +53,33 @@
...
-  static Counter counterEventSizeThresholdLevel1;
-  static Counter counterEventSizeThresholdLevel2;
-  static Counter counterEventSizeThresholdLevel3;
-  static Counter counterEventSizeThresholdLevel4;
-  static Counter counterTotalReadDataSize;
-  static Counter counterTotalReadTime;
-  static Counter counterTotalUploadSize;
-  static Counter counterTotalUploadTimeS;
-  static Counter counterTotalDroppedEventNum;
+  public static Counter counterTotalReceivedDataSize;
```
jerqi, line 56: why do we make these variable public?

storage/src/main/java/com/tencent/rss/storage/handler/impl/ComposedClientReadHandler.java
```diff
@@ -0,0 +1,109 @@
...
+  private static final int LEVEL2 = 2;
```
jerqi, line 34: hot warm cold freeze

storage/src/main/java/com/tencent/rss/storage/util/StorageType.java
```diff
@@ -22,5 +22,6 @@
   HDFS,
   LOCALFILE,
   FILE,
-  LOCALFILE_AND_HDFS
+  LOCALFILE_AND_HDFS,
+  MEMORY
```
jerqi, line 26: MEMORY shouldn't be a single storage

common/src/main/java/com/tencent/rss/common/exception/IndexFileLostException.java
```diff
@@ -0,0 +1,27 @@
...
+ * https://opensource.org/licenses/Apache-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package com.tencent.rss.common.exception;
+
```
jerqi, line 20: If we have block check, we don't need this.

server/src/main/java/com/tencent/rss/server/ShuffleServer.java
```diff
@@ -156,7 +160,14 @@ private void initialization() throws Exception {
...
     registerHeartBeat = new RegisterHeartBeat(this);
     shuffleFlushManager = new ShuffleFlushManager(shuffleServerConf, id, this, multiStorageManager);
-    shuffleBufferManager = new ShuffleBufferManager(shuffleServerConf, shuffleFlushManager);
+    memoryShuffleEnabled = shuffleServerConf.getBoolean(
+        ShuffleServerConf.SERVER_MEMORY_SHUFFLE_ENABLED);
+    if (memoryShuffleEnabled) {
+      shuffleBufferManager = new MemoryFSShuffleBufferManager(
+          shuffleServerConf, shuffleFlushManager);
+    } else {
+      shuffleBufferManager = new FSShuffleBufferManager(shuffleServerConf, shuffleFlushManager);
```
jerqi, line 169: We can unify FSShuffleBufferManaer and MemoryFSShuffleBufferManager.

server/src/main/java/com/tencent/rss/server/ShuffleServer.java
```diff
@@ -276,4 +287,8 @@ public boolean isHealthy() {
   public GRPCMetrics getGrpcMetrics() {
     return grpcMetrics;
   }
+
+  public boolean isMemoryShuffleEnabled() {
+    return memoryShuffleEnabled;
```
jerqi, line 292: check StorageType

storage/src/main/java/com/tencent/rss/storage/handler/impl/ComposedClientReadHandler.java
frankliee: recommended  level name:
hot
warm
cold
frozen or backup

storage/src/main/java/com/tencent/rss/storage/factory/ShuffleHandlerFactory.java
```diff
@@ -56,7 +58,28 @@ public static synchronized ShuffleHandlerFactory getInstance() {
...
+    return new ComposedClientReadHandler(handlers);
+  }
+
   public ClientReadHandler createShuffleReadHandler(CreateShuffleReadHandlerRequest request) {
+    String storageType = request.getStorageType();
+    String[] types = storageType.split(",");
+    if (types.length > 1) {
+      return createComposedReadHandler(types, request);
+    } else {
+      return createSingleReadHandler(request);
```
jerqi, line 78: Seemly we don't need SingleReadHandler

storage/src/main/java/com/tencent/rss/storage/handler/impl/ComposedClientReadHandler.java
frankliee: freeze is verb

storage/src/main/java/com/tencent/rss/storage/handler/impl/MemoryClientReadHandler.java
```diff
@@ -0,0 +1,98 @@
...
+  @Override
+  public ShuffleDataResult readShuffleData() {
+    boolean readSuccessful = false;
+    ShuffleDataResult result = null;
+
+    RssGetInMemoryShuffleDataRequest request = new RssGetInMemoryShuffleDataRequest(
+        appId,shuffleId, partitionId, lastBlockId, readBufferSize);
+
+    for (ShuffleServerClient shuffleServerClient : shuffleServerClients) {
+      if (excludeServerClients.contains(shuffleServerClient)) {
```
jerqi, line 64: remove excludeServerClients.

storage/src/main/java/com/tencent/rss/storage/handler/impl/MemoryClientReadHandler.java
jerqi, line 34: Not FileReader.

server/src/main/java/com/tencent/rss/server/ShuffleServerGrpcService.java
```diff
@@ -225,19 +231,23 @@ public void finishShuffle(FinishShuffleRequest req,
     int shuffleId = req.getShuffleId();
     StatusCode status;
     String msg = "OK";
-    String errorMsg = "Fail to finish shuffle for appId["
-        + appId + "], shuffleId[" + shuffleId + "], data may be lost";
-    try {
-      LOG.info("Get finishShuffle request for appId[" + appId + "], shuffleId[" + shuffleId + "]");
-      status = shuffleServer.getShuffleTaskManager().commitShuffle(appId, shuffleId);
-      if (status != StatusCode.SUCCESS) {
+    if (shuffleServer.isMemoryShuffleEnabled()) {
```
jerqi, line 234: Memory Client not commit, Server can also process normal client.

storage/src/main/java/com/tencent/rss/storage/handler/impl/ComposedClientReadHandler.java
colinmjj: ok

proto/src/main/proto/Rss.proto
colinmjj: To avoid misunderstanding, will update getShuffleIndex, getShuffleData -> getLocalShuffleIndex, getLocalShuffleData

server/src/main/java/com/tencent/rss/server/ShuffleServerMetrics.java
colinmjj: new package is created for buffer related class, these variable have to be public to access

storage/src/main/java/com/tencent/rss/storage/handler/impl/ComposedClientReadHandler.java
jerqi: frozen

storage/src/main/java/com/tencent/rss/storage/util/StorageType.java
colinmjj: will update storageType with possible combination

storage/src/main/java/com/tencent/rss/storage/handler/impl/ComposedClientReadHandler.java
colinmjj: hot, warm, cold will be used for now, frozen can be added later if necessary

common/src/main/java/com/tencent/rss/common/exception/IndexFileLostException.java
colinmjj: ok

server/src/main/java/com/tencent/rss/server/ShuffleServer.java
colinmjj: ok, keep only BufferManager to support both Local and Memory_Local

server/src/main/java/com/tencent/rss/server/ShuffleServer.java
colinmjj: This flag is not necessary, remove it

storage/src/main/java/com/tencent/rss/storage/factory/ShuffleHandlerFactory.java
colinmjj: There has no SingleReadHandler, this method will be refactor

storage/src/main/java/com/tencent/rss/storage/handler/impl/MemoryClientReadHandler.java
colinmjj: ok

server/src/main/java/com/tencent/rss/server/ShuffleServerGrpcService.java
colinmjj: ok,  client should decide if send commit rpc to shuffle server

client-spark/spark2/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
```diff
@@ -120,6 +123,17 @@ public RssShuffleWriter(
     this.shuffleWriteClient = shuffleWriteClient;
     this.shuffleServersForData = rssHandle.getShuffleServersForData();
     this.partitionToServers = rssHandle.getPartitionToServers();
+    setMemoryShuffleEnabled(sparkConf.get(RssClientConfig.RSS_STORAGE_TYPE));
+  }
+
+  private void setMemoryShuffleEnabled(String storageType) {
```
jerqi, line 129: isMemoryShuffleEnabled = getMemoryShuffleEnabled(storageType); 
may be better?

client-spark/spark3/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
```diff
@@ -124,6 +126,17 @@ public RssShuffleWriter(
     this.partitionLengths = new long[partitioner.numPartitions()];
     Arrays.fill(partitionLengths, 0);
     partitionToServers = rssHandle.getPartitionToServers();
+    setMemoryShuffleEnabled(sparkConf.get(RssClientConfig.RSS_STORAGE_TYPE));
```
jerqi, line 129: ditto

integration-test/spark-common/src/test/java/com/tencent/rss/test/SparkClientWithLocalTest.java
```diff
@@ -50,6 +43,14 @@
 import org.roaringbitmap.longlong.LongIterator;
 import org.roaringbitmap.longlong.Roaring64NavigableMap;
 
+import java.io.File;
```
jerqi, line 46: Need we a mores strict `import  order` in code style checker?

client-spark/spark2/src/main/java/org/apache/spark/shuffle/writer/RssShuffleWriter.java
colinmjj: ok

server/src/main/java/com/tencent/rss/server/ShuffleServerConf.java
```diff
@@ -36,18 +36,6 @@
...
 
-  public static final ConfigOption<Long> SERVER_BUFFER_SPILL_THRESHOLD = ConfigOptions
-      .key("rss.server.buffer.spill.threshold")
-      .longType()
-      .noDefaultValue()
-      .withDescription("Spill threshold for buffer manager, it must be less than rss.server.buffer.capacity");
-
-  public static final ConfigOption<Long> SERVER_PARTITION_BUFFER_SIZE = ConfigOptions
-      .key("rss.server.partition.buffer.size")
-      .longType()
```
jerqi, line 47: nit: checkValue. We have positiveLongValidator

integration-test/spark-common/src/test/java/com/tencent/rss/test/SparkClientWithLocalTest.java
colinmjj: The rule of import order should be added to project configuration.

server/src/main/java/com/tencent/rss/server/ShuffleServerConf.java
```diff
@@ -36,18 +36,6 @@
       .noDefaultValue()
       .withDescription("Max memory of buffer manager for shuffle server");
 
-  public static final ConfigOption<Long> SERVER_BUFFER_SPILL_THRESHOLD = ConfigOptions
-      .key("rss.server.buffer.spill.threshold")
-      .longType()
```
jerqi, line 41: nit: checkValue. We have positiveLongValidator

server/src/main/java/com/tencent/rss/server/ShuffleServerGrpcService.java
```diff
@@ -466,11 +476,78 @@ public void getShuffleIndex(GetShuffleIndexRequest request,
...
+    int readBufferSize = request.getReadBufferSize();
+    long start = System.currentTimeMillis();
+    StatusCode status = StatusCode.SUCCESS;
+    String msg = "OK";
+    GetMemoryShuffleDataResponse reply;
+    String requestInfo = "appId[" + appId + "], shuffleId[" + shuffleId + "], partitionId["
+        + partitionId + "]";
+
+    // todo: if can get the exact memory size?
+    if (shuffleServer.getShuffleBufferManager().requireReadMemoryWithRetry(readBufferSize)) {
```
jerqi, line 504: why we use read memory? i think it occupy write memory.

server/src/main/java/com/tencent/rss/server/ShuffleServerConf.java
colinmjj: > nit: checkValue. We have positiveLongValidator
the config is removed

server/src/main/java/com/tencent/rss/server/ShuffleServerGrpcService.java
colinmjj: when get data from shuffle server, always request memory first to avoid OOM, eg,
1. Request 10M data from memory
2. Get Data and it is flushed before send to client
3. Data is send to client
Between step 2&3, there has no any metrics to track 10M, it should be kind of memory leak.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,374 @@
...
+
+  // 1. generate buffer segments and other info: if blockId exist, start with which eventId
+  // 2. according to info from step 1, generate data
+  public synchronized ShuffleDataResult getShuffleData(
+      long lastBlockId, int readBufferSize) {
+    try {
+      List<BufferSegment> bufferSegments = Lists.newArrayList();
+      ActualReadInfo actualReadInfo = updateBufferSegments(
+          lastBlockId, readBufferSize, bufferSegments);
+      if (actualReadInfo.actualReadSize > 0) {
```
jerqi, line 149: getter may be better？

colinmjj: ok

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,374 @@
...
+  // here is the rule to read data in memory:
+  // 1. read from inFlushBlockMap order by eventId asc, then from blocks
+  // 2. if can't find lastBlockId, means related data may be flushed to storage, repeat step 1
+  private ActualReadInfo updateBufferSegments(
+      long lastBlockId,
+      long readBufferSize,
+      List<BufferSegment> bufferSegments) throws Exception {
+    long fromEventId = INVALID_EVENT_ID;
+    long currentBlockId = lastBlockId;
+    List<Long> sortedEventId = sortFlushingEventId();
```
jerqi, line 170: may we needn't sort the events.

colinmjj: Sort events to make sure the read order for bufferInCache/bufferInFlush is absolute consistent every time.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,374 @@
...
+      // find lastBlockId, then read from next block
+      boolean foundBlockId = false;
+      for (ShufflePartitionedBlock block : cachedBlocks) {
+        if (!foundBlockId) {
+          if (block.getBlockId() == lastBlockId) {
+            foundBlockId = true;
+          }
+          continue;
+        }
+        if (parameter instanceof byte[]) {
```
jerqi, line 293: Werid.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,374 @@
...
+    } catch (Exception e) {
+      LOG.error("Exception happened when getShuffleData in buffer", e);
+    }
+    return new ShuffleDataResult();
+  }
+
+  // here is the rule to read data in memory:
+  // 1. read from inFlushBlockMap order by eventId asc, then from blocks
+  // 2. if can't find lastBlockId, means related data may be flushed to storage, repeat step 1
+  private ActualReadInfo updateBufferSegments(
```
jerqi, line 164: This method is not easy to understand.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,374 @@
...
+    long fromEventId = INVALID_EVENT_ID;
+    long currentBlockId = lastBlockId;
+    List<Long> sortedEventId = sortFlushingEventId();
+    int offset = 0;
+    int eventIndex = 0;
+    long updatedLastBlockId = lastBlockId;
+    boolean hasBlockId = false;
+    // read from inFlushBlockMap first
+    if (inFlushBlockMap.size() > 0) {
+      for (Long eventId : sortedEventId) {
```
jerqi, line 177: if event != INVALID_EVENT_ID, eventId = fromEventId ?

storage/src/main/java/com/tencent/rss/storage/handler/impl/ComposedClientReadHandler.java
```diff
@@ -0,0 +1,109 @@
...
+public class ComposedClientReadHandler implements ClientReadHandler {
+
+  private static final Logger LOG = LoggerFactory.getLogger(ComposedClientReadHandler.class);
+
+  private ClientReadHandler hotDataReadHandler;
+  private ClientReadHandler warmDataReadHandler;
+  private ClientReadHandler coldDataReadHandler;
+  private static final int HOT = 1;
+  private static final int WARM = 2;
+  private static final int COLD = 3;
```
jerqi, line 35: No frozen handler which read upload handler?

storage/src/main/java/com/tencent/rss/storage/util/StorageType.java
```diff
@@ -21,6 +21,8 @@
 public enum StorageType {
   HDFS,
   LOCALFILE,
-  FILE,
-  LOCALFILE_AND_HDFS
+  LOCALFILE_AND_HDFS,
```
jerqi, line 24: LOCALFILE_HDFS consistent with others.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBufferManager.java
```diff
@@ -16,48 +16,65 @@
...
+  protected AtomicLong preAllocatedSize = new AtomicLong(0L);
+  protected AtomicLong inFlushSize = new AtomicLong(0L);
+  protected AtomicLong usedMemory = new AtomicLong(0L);
   private AtomicLong readDataMemory = new AtomicLong(0L);
   // appId -> shuffleId -> partitionId -> ShuffleBuffer to avoid too many appId
-  private Map<String, Map<Integer, RangeMap<Integer, ShuffleBuffer>>> bufferPool;
-
+  protected Map<String, Map<Integer, RangeMap<Integer, ShuffleBuffer>>> bufferPool;
+  // appId -> shuffleId -> shuffle size in buffer
+  protected Map<String, Map<Integer, AtomicLong>> shuffleSizeMap = Maps.newConcurrentMap();
```
jerqi, line 66: It seems not to have clean up logic.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBufferManager.java
```diff
@@ -293,14 +357,68 @@ public long getCapacity() {
...
+    return pickedShuffle;
+  }
+
+  private List<Map.Entry<String, AtomicLong>> generateSizeList() {
+    Map<String, AtomicLong> sizeMap = Maps.newHashMap();
+    for (Map.Entry<String, Map<Integer, AtomicLong>> appEntry : shuffleSizeMap.entrySet()) {
+      String appId = appEntry.getKey();
+      for (Map.Entry<Integer, AtomicLong> shuffleEntry : appEntry.getValue().entrySet()) {
+        Integer shuffleId = shuffleEntry.getKey();
+        sizeMap.put(appId + KEY_SPLIT_CHAR + shuffleId, shuffleEntry.getValue());
```
jerqi, line 410: RssUtils.generateShuffleKey

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,374 @@
...
+
+    return mSize;
+  }
+
+  public synchronized ShuffleDataFlushEvent toFlushEvent(
+      String appId,
+      int shuffleId,
+      int startPartition,
+      int endPartition,
+      Map<String, Map<Integer, AtomicLong>> shuffleSizeMap,
```
jerqi, line 80: I think we should only pass a size not a map.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: Kind of weird, it's used to avoid duplicated code

server/src/main/java/com/tencent/rss/server/ShuffleServerGrpcService.java
```diff
@@ -162,6 +166,7 @@ public void sendShuffleData(SendShuffleDataRequest req,
           String errorMsg = "Error happened when shuffleEngine.write for "
               + shuffleDataInfo + ": " + e.getMessage();
           ret = StatusCode.INTERNAL_ERROR;
+          e.printStackTrace();
```
jerqi, line 169: use log.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
jerqi: I worry the the time cost of the sort.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
jerqi: I think we can extract some common methods  to avoid duplicated code.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,374 @@
...
+    CopyShuffleDataCall(byte[] data, ShufflePartitionedBlock block, int offset) {
+      this.block = block;
+      this.data = data;
+      this.offset = offset;
+    }
+
+    @Override
+    public Integer call() throws Exception {
+      try {
+        System.arraycopy(block.getData(), 0, data, offset, block.getLength());
```
jerqi, line 321: copy is necessary?

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,374 @@
...
+      long lastBlockId,
+      long readBufferSize,
+      List<BufferSegment> bufferSegments) throws Exception {
+    long fromEventId = INVALID_EVENT_ID;
+    long currentBlockId = lastBlockId;
+    List<Long> sortedEventId = sortFlushingEventId();
+    int offset = 0;
+    int eventIndex = 0;
+    long updatedLastBlockId = lastBlockId;
+    boolean hasBlockId = false;
```
jerqi, line 174: If block is flushed,  we need to read from the start of buffer. Is possible to read duplicated data.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,374 @@
...
+    // read from inFlushBlockMap first
```
jerqi, line 175: comments should explain why inFlushBlockMap first.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: The method is target to update `List<BufferSegment> bufferSegments`, and also update read info eg, lastBlockId

storage/src/main/java/com/tencent/rss/storage/handler/impl/ComposedClientReadHandler.java
colinmjj: Currently, there is MultiStorageReadHandler for both local and hdfs, for memory + MultiStorageReadHandler, only hot & warm will be created.

storage/src/main/java/com/tencent/rss/storage/util/StorageType.java
colinmjj: ok

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBufferManager.java
colinmjj: It's the only place using key as `appId + KEY_SPLIT_CHAR + shuffleId`, it can be wrapped as a method, but shouldn't be added to RssUtils until it is called from other place.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: The map is used to track shuffleId -> dataSize, it should be updated when the size is changed.
ShuffleBuffer has to get this map, and pass it as a parameter is one of ways.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBufferManager.java
jerqi: We use appId + / + shuffleId as shufflekey in mulitstorage code.

server/src/main/java/com/tencent/rss/server/ShuffleServerGrpcService.java
colinmjj: it should be removed, good catch

storage/src/main/java/com/tencent/rss/storage/handler/impl/ComposedClientReadHandler.java
jerqi: OK, @frankliee If you have time, please implement this function.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: yes, it's the transform like  `List<byte[]> -> byte[]`

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
jerqi: so the method name is weird. The name can't reflect the behavior of the method.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: yes, it will cause a little performance problem, but it's ok for the first step

integration-test/spark-common/src/test/java/com/tencent/rss/test/SparkClientWithLocalTest.java
jerqi: @frankliee Can you add these rules?

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: ok

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
jerqi: At least, we should left some `to do` or comments.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
jerqi: One Buffer only care the shuffle size of itself.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: ok

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,374 @@
...
+      return new ActualReadInfo(offset, fromEventId, updatedLastBlockId);
+    }
+  }
+
+  private void updateShuffleData(
+      long lastBlockId,
+      long readBufferSize,
+      byte[] data,
+      long fromEventId) throws Exception {
+    List<Long> sortedEventId = sortFlushingEventId();
```
jerqi, line 225: We sort events multi times. If event is flushed. The views will be different.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: ok, size update can be called after `buffer.toFlushEvent` in `ShuffleBufferManager`

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
jerqi: And log some necessary information.

integration-test/spark-common/src/test/java/com/tencent/rss/test/SparkClientWithLocalTest.java
frankliee: I will do it.

storage/src/main/java/com/tencent/rss/storage/util/StorageType.java
colinmjj: I'm prefer to update this in another PR

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,374 @@
...
+
+    ActualReadInfo(int actualReadSize, long fromEventId, long lastBlockId) {
+      this.actualReadSize = actualReadSize;
+      this.fromEventId = fromEventId;
+      this.lastBlockId = lastBlockId;
+    }
+  }
+
+  class ListReadInfo {
+    int offset;
```
jerqi, line 366: What's meaning of offset?

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: ok, method name can be changed as updateBufferSegmentsAndGetReadInfo

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: Actually, here is to avoid duplicated workflow, or flat the code for better readable.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: With properly configuration,  inFlushBlockMap won't have much element. Every element means flushing, I think there has some problem with storage if it has 10+ elements. So, it won't cost much time.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,374 @@
...
+      List<BufferSegment> bufferSegments) throws Exception {
+    long fromEventId = INVALID_EVENT_ID;
+    long currentBlockId = lastBlockId;
+    List<Long> sortedEventId = sortFlushingEventId();
+    int offset = 0;
+    int eventIndex = 0;
+    long updatedLastBlockId = lastBlockId;
+    boolean hasBlockId = false;
+    // read from inFlushBlockMap first
+    if (inFlushBlockMap.size() > 0) {
```
jerqi, line 176: This process is a little complex.I feel whether we should add an interface getAllBlockId, and readData with the list  of blockIds in the client.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: It's ok, read won't depend on the view status, just get data according to view & blockId. The backward is duplicated data may be read.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: The offset is used to update the offset of segment

storage/src/main/java/com/tencent/rss/storage/util/StorageType.java
jerqi: OK.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
jerqi: I feel readability is more important.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
jerqi: OK.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: I don't think the process will be simplifier with such interface.
The current process is:
1. get the start point, some cases here
2. update segment from start point
3. fill data from start point
4. return

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
jerqi: Actually, i don't understand what BufferSegment is.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
jerqi: If we know the blockIds that we need to read, we only travel all the shuffleBuffers once, we can get all the blocks.

jerqi: We don't need sort. And find block multi times.

colinmjj: if all blockIds is flushed, but there still data in memory, return 0?

integration-test/spark-common/src/test/java/com/tencent/rss/test/SparkClientWithLocalTest.java
duanmeng: Use iceberg rule https://github.com/apache/iceberg/blob/master/.baseline/checkstyle/checkstyle.xml#L247 @frankliee

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,374 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Supplier;
+
+public class ShuffleBuffer {
```
duanmeng, line 41: Could we break the list of `ShufflePartitionBlocks` into index buffer and data buffer since we support reading directly from memory?

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,374 @@
...
+    return inFlushBlockMap;
+  }
+
+  // 1. generate buffer segments and other info: if blockId exist, start with which eventId
+  // 2. according to info from step 1, generate data
+  public synchronized ShuffleDataResult getShuffleData(
+      long lastBlockId, int readBufferSize) {
+    try {
+      List<BufferSegment> bufferSegments = Lists.newArrayList();
+      ActualReadInfo actualReadInfo = updateBufferSegments(
```
jerqi, line 147: BufferSegment should be a field of ActualReadInfo.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,374 @@
...
+      long lastBlockId,
+      long readBufferSize,
+      Object parameter) throws Exception {
+    int currentOffset = offset;
+    Callable<Integer> callable = null;
+    if (lastBlockId == Constants.INVALID_BLOCK_ID) {
+      // read from first block
+      for (ShufflePartitionedBlock block : cachedBlocks) {
+        // do action
+        if (parameter instanceof byte[]) {
```
duanmeng, line 271: It may be possible and appropirate to use visitor pattern and double dispatch to avoid these instanceof matching.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,374 @@
...
+  }
+
+  // 1. generate buffer segments and other info: if blockId exist, start with which eventId
+  // 2. according to info from step 1, generate data
+  public synchronized ShuffleDataResult getShuffleData(
+      long lastBlockId, int readBufferSize) {
+    try {
+      List<BufferSegment> bufferSegments = Lists.newArrayList();
+      ActualReadInfo actualReadInfo = updateBufferSegments(
+          lastBlockId, readBufferSize, bufferSegments);
```
jerqi, line 148: updateBufferSegments -> calculateActualReadInfo

storage/src/main/java/com/tencent/rss/storage/handler/impl/MemoryClientReadHandler.java
```diff
@@ -0,0 +1,89 @@
...
+        RssGetInMemoryShuffleDataResponse response =
+            shuffleServerClient.getInMemoryShuffleData(request);
+        result = new ShuffleDataResult(response.getData(), response.getBufferSegments());
+
+        readSuccessful = true;
+        break;
+      } catch (Exception e) {
+        // todo: fault tolerance solution should be added
+        LOG.warn("Failed to read in memory shuffle data with "
+            + shuffleServerClient.getClientInfo(), e);
```
duanmeng, line 71: It'd better to use placeholder. `LOG.warn("Failed to read in memory shuffle data with "
            + shuffleServerClient.getClientInfo(), e);` -> `LOG.warn("Failed to read in memory shuffle data with {}"
            ,shuffleServerClient.getClientInfo(), e)`

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: ok

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: it will be removed for better readable

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: ok

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
jerqi, line 176: isEmpty is better

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: there will be a lot of System.arrayCopy which is an expensive call

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,374 @@
...
+    long currentBlockId = lastBlockId;
+    List<Long> sortedEventId = sortFlushingEventId();
+    int offset = 0;
+    int eventIndex = 0;
+    long updatedLastBlockId = lastBlockId;
+    boolean hasBlockId = false;
+    // read from inFlushBlockMap first
+    if (inFlushBlockMap.size() > 0) {
+      for (Long eventId : sortedEventId) {
+        ListReadInfo listReadInfo = processCachedBlocks(offset, inFlushBlockMap.get(eventId),
```
jerqi, line 178: `ListReadInfo` need better name

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: ok

storage/src/main/java/com/tencent/rss/storage/handler/impl/MemoryClientReadHandler.java
colinmjj: ok

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
jerqi: > It may be possible and appropirate to use visitor pattern and double dispatch to avoid these instanceof matching.

+1, Could we extract the way that visit block.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: how about
```
class CachedBlocksReadInfo {
    int offsetInResultData;
    boolean hasBlockId;
```

jerqi: Good !. But could`hasBlockId`  have better name.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: Readable is important, will remove such trick code

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: ok, for readable, maybe `foundBlockIdInBlocks`

jerqi: OK

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
jerqi: Actually ActualReadInfo should have a field which type is List<ShufflePartitionedBlock>. This field will be updated when buffersegment is updated. And we needn't found any block, when we copy data.

integration-test/spark-common/src/test/java/com/tencent/rss/test/SparkClientWithLocalTest.java
frankliee: Iceberg's style is wired.

import com.tencent.rss.common.Arguments;
import com.tencent.rss.common.metrics.GRPCMetrics;
import com.tencent.rss.common.metrics.JvmMetrics;
import com.tencent.rss.common.rpc.ServerInterface;
import com.tencent.rss.common.web.CommonMetricsServlet;
import com.tencent.rss.common.web.JettyServer;
import io.prometheus.client.CollectorRegistry;
import java.io.FileNotFoundException;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
import picocli.CommandLine;

frankliee: In fact, spark disables its java import checker....

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: ActualReadInfo will be removed

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,285 @@
...
+    eventIdList.sort((id1, id2) -> {
+      if (id1 > id2) {
+        return 1;
+      }
+      return -1;
+    });
+    return eventIdList;
+  }
+
+  private CachedBlocksReadInfo updateBufferSegmentsWithCachedBlocks(
```
jerqi, line 219: better name？

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,285 @@
...
+    CachedBlocksReadInfo(int offsetInResultData, boolean hasLastBlockId) {
+      this.offsetInResultData = offsetInResultData;
+      this.hasLastBlockId = hasLastBlockId;
+    }
+
+    int getOffsetInResultData() {
+      return offsetInResultData;
+    }
+
+    boolean isHasLastBlockId() {
```
jerqi, line 280: remove `is`

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,285 @@
...
+      long lastBlockId,
+      long readBufferSize,
+      List<BufferSegment> bufferSegments,
+      List<ShufflePartitionedBlock> resultBlocks) {
+    long nextBlockId = lastBlockId;
+    List<Long> sortedEventId = sortFlushingEventId();
+    int offset = 0;
+    boolean hasLastBlockId = false;
+    // read from inFlushBlockMap first to make sure the order of
+    // data read is according to the order of data received
```
jerqi, line 158: can't guarantee the order, just try our best?

jerqi: If a block is in the cacheBlocks when we seek flushBlocks, and the block is the flushBlocks when we read cache blocks. Seemly we can't read the block.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: ok

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,285 @@
...
+    if (!inFlushBlockMap.isEmpty()) {
+      for (Long eventId : sortedEventId) {
```
jerqi, line 160: Add comments: `// The number of events should be less than 10.`

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,285 @@
...
+        // update bufferSegments with current blocks in flush map
+        CachedBlocksReadInfo cachedBlocksReadInfo = updateBufferSegmentsWithCachedBlocks(
+            offset, inFlushBlockMap.get(eventId), nextBlockId, readBufferSize,
+            bufferSegments, resultBlocks);
+        offset = cachedBlocksReadInfo.getOffsetInResultData();
+        // if has data or find last blockId, read from begin with next cached blocks
+        if (offset > 0 || cachedBlocksReadInfo.isHasLastBlockId()) {
```
jerqi, line 167: We judge hasLastBlockId is enough?

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: There has a lock to make sure it won't happen

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
```diff
@@ -0,0 +1,285 @@
...
+      }
+    }
+    // try to read from cached blocks which is not in flush queue
+    if (blocks.size() > 0 && offset < readBufferSize) {
+      CachedBlocksReadInfo cachedBlocksReadInfo = updateBufferSegmentsWithCachedBlocks(
+          offset, blocks, nextBlockId, readBufferSize, bufferSegments, resultBlocks);
+      offset = cachedBlocksReadInfo.getOffsetInResultData();
+      hasLastBlockId = cachedBlocksReadInfo.isHasLastBlockId();
+    }
+    if (lastBlockId != Constants.INVALID_BLOCK_ID && offset == 0 && !hasLastBlockId) {
```
jerqi, line 184: `offset == 0` is enough?

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: > 

data in eventId 1 must arrive in Shffle server early than data in eventId 2.
data in flush must arrive in  Shffle server early than data in cached blocs.
Guarantee the order by received time

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: ok

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: offset > 0 is deal with lastBlockId = -1, it can be simplifier to keep hasLastBlockId only.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: offset = 0 means there has no data, if lastBlockId != -1,  here is the different case:
1. can't find lastBlockId, we should read from begin
2. find lastBlockId, it is the last block, just return empty

jerqi: OK

storage/src/main/java/com/tencent/rss/storage/handler/impl/MemoryClientReadHandler.java
jerqi: remind this comment @colinmjj

colinmjj: how about change name `AbstractFileClientReadHandler` -> `AbstractClientReadHandler`

jerqi: OK

storage/src/main/java/com/tencent/rss/storage/handler/impl/ComposedClientReadHandler.java
frankliee: I will handle it.

jerqi:
All my suggestions are  above.

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: `updateBufferSegmentsAndResultBlocks`?

jerqi: OK, CacheBlocksReadInfo should be a paramter?

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
jerqi: remove `is`

server/src/main/java/com/tencent/rss/server/ShuffleServerConf.java
```diff
@@ -308,6 +296,28 @@
       .defaultValue("com.tencent.rss.server.LocalStorageChecker")
       .withDescription("The list of the Checker's name");
 
+  public static final ConfigOption<Boolean> SERVER_MEMORY_SHUFFLE_ENABLED = ConfigOptions
```
jerqi, line 299: remove?

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
duanmeng: OK

storage/src/main/java/com/tencent/rss/storage/handler/impl/MemoryClientReadHandler.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    if (!readSuccessful) {
+      throw new RssException("Failed to read in memory shuffle data for appId[" + appId
+          + "], shuffleId[" + shuffleId + "], partitionId[" + partitionId + "]");
+    }
+
+    // update lastBlockId for next rpc call
+    byte[] data = result.getData();
+    if (data != null && data.length > 0) {
```
duanmeng, line 82: It's more safe and convenient to use `if (!result.isEmpty())`

server/src/main/java/com/tencent/rss/server/ShuffleServerConf.java
colinmjj: ok

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: ok

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
colinmjj: The current name is `readCachedBlocksAndUpdateSegments`

storage/src/main/java/com/tencent/rss/storage/handler/impl/MemoryClientReadHandler.java
colinmjj: ok

server/src/main/java/com/tencent/rss/server/buffer/ShuffleBuffer.java
jerqi: separate two methods.

colinmjj: ok

jerqi:
LGTM

jerqi (approved):

colinmjj:
@jerqi @duanmeng @frankliee thanks for review

PR #39, opened 2021-12-17
Check blockIds in read client to allow early return

frankliee:
Add a pendingBlockIds to track all blocks should be waited.

Check blockIds in read client to allow early return

No.

W/O checkBlockIds (1000 partitions): 9 ms
W/  checkBlockIds (1000 partitions): 31 ms

W/O checkBlockIds (10000 partitions): 24 ms
W/  checkBlockIds (10000 partitions): 93 ms

W/O checkBlockIds (100000 partitions): 163 ms
W/  checkBlockIds (100000 partitions): 582 ms


client/src/main/java/com/tencent/rss/client/impl/ShuffleReadClientImpl.java
```diff
@@ -122,6 +131,12 @@ public CompressedShuffleBlock readShuffleBlockData() {
         return null;
       }
     }
+
+    // All blocks are processed, so just return
+    if (pendingBlockIds.isEmpty()) {
```
colinmjj, line 136: The check should be called before `if (bufferSegmentQueue.isEmpty()) {`

client/src/test/java/com/tencent/rss/client/impl/CheckBlockIdsBenchmark.java
```diff
@@ -0,0 +1,80 @@
...
+import java.io.IOException;
+import java.util.Iterator;
+import java.util.concurrent.atomic.AtomicLong;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.roaringbitmap.longlong.Roaring64NavigableMap;
+import com.tencent.rss.client.util.ClientUtils;
+
+public class CheckBlockIdsBenchmark extends HdfsTestBase {
```
colinmjj, line 33: it shouldn't extends HdfsTestBase

colinmjj:
There should be a new TestCase, for example:
expected blockId: 1,2,3,4
actual blockId in storage:  1, 2, 3, 4, 5, 6
blockId 5 & 6 shouldn't be read.

client/src/test/java/com/tencent/rss/client/impl/CheckBlockIdsBenchmark.java
```diff
@@ -0,0 +1,80 @@
...
+  private static final String EXPECTED_EXCEPTION_MESSAGE = "Exception should be thrown";
+  private static AtomicLong ATOMIC_LONG = new AtomicLong(0);
+  @Rule
+  public ExpectedException thrown = ExpectedException.none();
+
+  @Test
```
colinmjj, line 39: It shouldn't be mark as regular test case, it can be executed manually.

client/src/main/java/com/tencent/rss/client/impl/ShuffleReadClientImpl.java
frankliee: Done

client/src/test/java/com/tencent/rss/client/impl/CheckBlockIdsBenchmark.java
frankliee: Fixed

client/src/test/java/com/tencent/rss/client/impl/CheckBlockIdsBenchmark.java
frankliee: Use normal main function instead.

frankliee:
> There should be a new TestCase, for example:
> expected blockId: 1,2,3,4
> actual blockId in storage: 1, 2, 3, 4, 5, 6
> blockId 5 & 6 shouldn't be read.

frankliee:
Add readTest15 @colinmjj

client/src/test/java/com/tencent/rss/client/impl/ShuffleReadClientImplTest.java
```diff
@@ -430,6 +430,32 @@ public void readTest14() throws Exception {
     readClient.close();
   }
 
+  @Test
+  public void readTest15() throws Exception {
```
colinmjj, line 434: The test is to filter unexpected taskId, it's not this PR's target

client/src/test/java/com/tencent/rss/client/impl/CheckBlockIdsBenchmark.java
```diff
@@ -0,0 +1,76 @@
...
+import java.io.IOException;
+import java.util.Iterator;
+import java.util.concurrent.atomic.AtomicLong;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.roaringbitmap.longlong.Roaring64NavigableMap;
+import com.tencent.rss.client.util.ClientUtils;
+
+public class CheckBlockIdsBenchmark {
```
duanmeng, line 33: It may not be appropriate to add a benchmark in the ut package especially for this single operation's benchmark. You could just add the conclusion in your normal codes comments.

frankliee: Spark seems to contain some benchmarks in code...

jerqi: I think microbenchmark is ok in the code. But our code style is not put benchmark into our code. We should keep consistent with the former code.

frankliee: Removed benchmark...

client/src/test/java/com/tencent/rss/client/impl/ShuffleReadClientImplTest.java
frankliee: Update: write 6 blocks, read 3 of these blocks.

colinmjj (approved):
LGTM

PR #47, opened 2021-12-25
Shutdown non-daemon executor service to avoid hang when jvm destroy

toujours33:
shut down non daemon executor service created in sendCommit by RssShuffleWriter

non daemon thread will block jvm destroy, which will lead to hang after spark job finished

No

run simple spark job in local mode with RSS, for example:
`./bin/run-example --master local GroupByTest 10 50000 4000`
before this patch, the command will never exist until we interrupt it
after this patch, normal exist after spark job finished


jerqi (approved):

jerqi:
@toujours33 Thanks for your contribution.

toujours33:
My pleasure

PR #50, opened 2021-12-27
Throw InterruptedException when report shuffle result

colinmjj:
Throw InterruptedException when report shuffle result, the exception is aways caused by speculation task killed

Shouldn't wrap it as RSSException, and InterruptedException should be handled by executor.

No

exist UT


client/src/main/java/com/tencent/rss/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -284,9 +284,13 @@ public void reportShuffleResult(
           break;
         }
       } catch (Exception e) {
```
jerqi, line 286: catch (InterruptedException ie) {

} catch(Exception e) {

}

colinmjj: Can't catch InterruptedException, because it is never thrown in try block.

jerqi: can catch.

colinmjj:
@jerqi The current fix is based on following exception trace
WeChatWorkScreenshot_514c96bd-9c8f-4506-bd3b-0ce5b134ea12

PR #52, opened 2021-12-28
Fix time unit from micro to mill

colinmjj:
Fix time unit from micro to mill

bug fix

No

exist UT


jerqi:
Why do we need this patch？

colinmjj:
@jerqi The time unit is incorrect, which cause too much check than expected, and too much log is generated
WeChatWorkScreenshot_4064a24c-a1d0-4e4f-8be0-a69d739f548e

jerqi:
Is that  all replaces that need to be fixed?

colinmjj:
> Is that all replaces that need to be fixed?

yes, replace all microSeconds with milliSeconds

jerqi (approved):

PR #59, opened 2022-01-04
[Improvement] Reduce memory cost in spark executor

colinmjj:
To reduce memory cost in Spark executor, the following improvements are included:
1. when shuffle read, don't copy byte[] for every block
2. when shuffle write, metadata should be cleared

To reduce memory and get better performance

No

Update exist UT

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/RssShuffleUtils.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/RssShuffleUtils.java
@@ -119,7 +119,7 @@ public class RssShuffleUtils {
   public static ByteBuffer decompressData(ByteBuffer data, int uncompressLength) {
     LZ4FastDecompressor fastDecompressor = LZ4Factory.fastestInstance().fastDecompressor();
     ByteBuffer uncompressData = ByteBuffer.allocateDirect(uncompressLength);
-    fastDecompressor.decompress(data, 0, uncompressData, 0, uncompressLength);
+    fastDecompressor.decompress(data, data.position(), uncompressData, 0, uncompressLength);
     return uncompressData;
   }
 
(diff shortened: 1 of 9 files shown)


client-spark/common/src/main/java/org/apache/spark/shuffle/RssShuffleUtils.java
```diff
@@ -119,7 +119,7 @@
   public static ByteBuffer decompressData(ByteBuffer data, int uncompressLength) {
     LZ4FastDecompressor fastDecompressor = LZ4Factory.fastestInstance().fastDecompressor();
     ByteBuffer uncompressData = ByteBuffer.allocateDirect(uncompressLength);
-    fastDecompressor.decompress(data, 0, uncompressData, 0, uncompressLength);
+    fastDecompressor.decompress(data, data.position(), uncompressData, 0, uncompressLength);
```
jerqi, line 122: why do we use `data.postition`?

colinmjj: position means the start point, it's the right value here. The 0 is a bug, it works because the start point is always 0 before this PR.

jerqi: What's the influence of this bug? Should we add some tests?

colinmjj: with previous implementation, if data.position != 0, the decompress will failed.
The current UT can catch the bug, it's ok without new UT.

jerqi: Spark application will fail sometimes？

client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -94,7 +94,7 @@ private void sendShuffleData(String taskId, List<ShuffleBlockInfo> shuffleDataIn
         for (ShuffleBlockInfo sbi : shuffleDataInfoList) {
           releaseSize += sbi.getFreeMemory();
         }
-        taskToBuffManager.get(taskId).freeAllocatedMemory(releaseSize);
+        taskToBufferManager.get(taskId).freeAllocatedMemory(releaseSize);
```
duanmeng, line 97: It'd better add existence checking to avoid NPE.

jerqi (approved):

client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
jerqi: Sorry, I miss the comment, but in our logic, we can guarantee this taskId's existence.

PR #56, opened 2021-12-30
[Feature] Support read from localfile, hdfs and uploaded hdfs

jerqi:
### What changes were proposed in this pull request?
We replace MultiStorageReadHandler with  ComposedClientReadHandler 

### Why are the changes needed?
#43 We can write hdfs and localfile. In this pull request, We unify the logic of read handler, and we can read three type data: data in localfile, data in hdfs, and data was uploaded. 

### Does this PR introduce _any_ user-facing change?
No

### How was this patch tested?
Add new ut.

PR #60, opened 2022-01-04
[Feature] Hdfs write fail, the data will write to localfile

jerqi:
When we fail to write hdfs, we use localfile instead in multistorage.

#56 we support mutlistorage read data, but hdfs is unstable sometimes. So we need a fallback mechanism that when data can't be written.

No

New UT


server/src/main/java/com/tencent/rss/server/ShuffleFlushManager.java
```diff
@@ -193,15 +192,29 @@ private void flushToFile(ShuffleDataFlushEvent event) {
...
           } catch (Exception e) {
             LOG.warn("Exception happened when write data for " + event + ", try again", e);
             ShuffleServerMetrics.counterWriteException.inc();
             Thread.sleep(1000);
           } finally {
             storage.unlockShuffleShared(RssUtils.generateShuffleKey(event.getAppId(), event.getShuffleId()));
           }
-          retry++;
-        }
+          event.increaseRetryTimes();
```
duanmeng, line 203: The fallback granularity is per event?

server/src/main/java/com/tencent/rss/server/ShuffleDataFlushEvent.java
```diff
@@ -35,6 +36,7 @@
   private List<ShufflePartitionedBlock> shuffleBlocks;
   private Supplier<Boolean> valid = null;
   private ShuffleBuffer shuffleBuffer;
+  private AtomicInteger retryTimes = new AtomicInteger();
```
duanmeng, line 39: final

server/src/main/java/com/tencent/rss/server/ShuffleFlushManager.java
jerqi: Yes

server/src/main/java/com/tencent/rss/server/ShuffleDataFlushEvent.java
jerqi: OK

colinmjj:
I think storage should has api `supportFallbak` as current implementation.
But how to fallback, it should be storage's action which should be wrapped in storage.
how about update interface `ShuffleWriteHandler.write(List)` -> 'ShuffleWriteHandler.write(Event)', the information like retry times can be got in write method.

jerqi:
> I think storage should has api `supportFallbak` as current implementation. But how to fallback, it should be storage's action which should be wrapped in storage. how about update interface `ShuffleWriteHandler.write(List)` -> 'ShuffleWriteHandler.write(Event)', the information like retry times can be got in write method.

you means that we should have a multistorageWriteHanlder to fallback?

jerqi:
> I think storage should has api `supportFallbak` as current implementation. But how to fallback, it should be storage's action which should be wrapped in storage. how about update interface `ShuffleWriteHandler.write(List)` -> 'ShuffleWriteHandler.write(Event)', the information like retry times can be got in write method.

Do you mean that Storage has api `supportFallback` instead StorageManager?

jerqi:
All comments are addressed.

jerqi:
GA failed because of flaky tests. And I rerun them.

coordinator/src/main/java/com/tencent/rss/coordinator/CoordinatorGrpcService.java
```diff
@@ -135,7 +135,7 @@ public void heartbeat(
         .setRetMsg("")
         .setStatus(StatusCode.SUCCESS)
         .build();
-    LOG.debug("Got heartbeat from " + serverNode);
+    LOG.warn("Got heartbeat from " + serverNode);
```
colinmjj, line 138: for 100+ shuffle server, there are too much heartbeat log

server/src/main/java/com/tencent/rss/server/ShuffleServerConf.java
```diff
@@ -336,6 +336,13 @@
       .defaultValue(64L * 1024L * 1024L)
       .withDescription("For multistorage, the event size exceed this value, flush data  to cold storage");
 
+  public static final ConfigOption<Long> FALLBACK_TIMES = ConfigOptions
+      .key("rss.server.fallback.times")
```
colinmjj, line 340: rss.server.multipleStorage.failedTimesBeforeFallback?

storage/src/main/java/com/tencent/rss/storage/common/Storage.java
```diff
@@ -42,6 +42,8 @@
 
   ShuffleWriteHandler getOrCreateWriteHandler(CreateShuffleWriteHandlerRequest request) throws IOException;
 
+  CreateShuffleWriteHandlerRequest getCreateWriterHandlerRequest(String appId, int shuffleId, int partition);
```
colinmjj, line 45: why this api is necessary?

jerqi: When we write hdfs failed, we need use the request to create local handler.

server/src/main/java/com/tencent/rss/server/ShuffleServerConf.java
jerqi: OK

coordinator/src/main/java/com/tencent/rss/coordinator/CoordinatorGrpcService.java
jerqi: Ok

server/src/main/java/com/tencent/rss/server/ShuffleServerConf.java
jerqi: For configuration, we should use lower case letters.

storage/src/main/java/com/tencent/rss/storage/common/Storage.java
colinmjj: ok

server/src/main/java/com/tencent/rss/server/ShuffleServerConf.java
colinmjj: ok

server/src/main/java/com/tencent/rss/server/ShuffleServerConf.java
```diff
@@ -336,6 +336,13 @@
       .defaultValue(64L * 1024L * 1024L)
       .withDescription("For multistorage, the event size exceed this value, flush data  to cold storage");
 
+  public static final ConfigOption<Long> FALLBACK_TIMES = ConfigOptions
```
colinmjj, line 339: FALLBACK_TIMES -> FALLBACK_MAX_FAIL_TIMES

jerqi: OK

colinmjj:
LGTM

colinmjj (approved):

PR #65, opened 2022-01-11
[Bug] LocalStorageChecker should support MEMORY_LOCALFILE and MEMORY_LOCALFILE_HDFS

jerqi:
Add a method containsLocalFile that judge storageType use localfile, and LOCALFILE, LOCALFILE_HDFS,MEMORY_LOCALFILE,MEMORY_LOCALFILE_HDFS will return true.

If we don't have this patch, MEMORY_LOCALFILE and MEMORY_LOCALFILE_HDFS can't use LocalStorageChecker.

NO

Manually.

### DIFF
--- a/common/src/main/java/com/tencent/rss/common/util/RssUtils.java
+++ b/common/src/main/java/com/tencent/rss/common/util/RssUtils.java
@@ -29,6 +29,7 @@ import java.io.InputStreamReader;
 import java.net.Inet4Address;
 import java.net.InetAddress;
 import java.net.NetworkInterface;
+import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.Enumeration;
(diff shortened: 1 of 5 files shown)


storage/src/main/java/com/tencent/rss/storage/util/ShuffleStorageUtils.java
```diff
@@ -239,4 +239,15 @@ public static long uploadFile(File file, HdfsFileWriter writer, int bufferSize)
       throw new IOException(e);
     }
   }
+
+  public static boolean containsLocalFile(String storageType) {
```
colinmjj, line 243: return StorageType.LOCALFILE.name().equals(storageType)
        || StorageType.LOCALFILE_HDFS.name().equals(storageType)
        || StorageType.LOCALFILE_HDFS_2.name().equals(storageType)
        || StorageType.MEMORY_LOCALFILE.name().equals(storageType)
        || StorageType.MEMORY_LOCALFILE_HDFS.name().equals(storageType);

jerqi: OK.

common/src/main/java/com/tencent/rss/common/util/RssUtils.java
```diff
@@ -161,29 +162,33 @@ public static Roaring64NavigableMap deserializeBitMap(byte[] bytes) throws IOExc
...
+        if (bufferOffset >= readBufferSize) {
+          ShuffleDataSegment sds = new ShuffleDataSegment(fileOffset, bufferOffset, bufferSegments);
+          dataFileSegments.add(sds);
+          bufferSegments = Lists.newArrayList();
+          bufferOffset = 0;
+          fileOffset = -1;
+        }
+      } catch (BufferUnderflowException ue) {
+        LOGGER.warn("Read index data under flow", ue);
+        break;
```
colinmjj, line 191: here is for bug fix with broken index data?

jerqi: Yes

jerqi: I separate them. It should be another pr.

colinmjj (approved):
LGTM

PR #67, opened 2022-01-11
[Bug] Support to read incomplete index data

jerqi:
When we encounter the incomplete index, we just stop reading, not throw an exception.

When we use the storageType MEMORY_HDFS, MEMORY_LOCALFILE_HDFS, after we read data from memory, the data may be flushed to HDFS. So we could read incomplete index data.

No

New ut

### DIFF
--- a/common/src/main/java/com/tencent/rss/common/util/RssUtils.java
+++ b/common/src/main/java/com/tencent/rss/common/util/RssUtils.java
@@ -29,6 +29,7 @@ import java.io.InputStreamReader;
 import java.net.Inet4Address;
 import java.net.InetAddress;
 import java.net.NetworkInterface;
+import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.Enumeration;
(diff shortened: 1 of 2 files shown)


common/src/main/java/com/tencent/rss/common/util/RssUtils.java
```diff
@@ -161,29 +162,33 @@ public static Roaring64NavigableMap deserializeBitMap(byte[] bytes) throws IOExc
...
-        fileOffset = -1;
+        if (bufferOffset >= readBufferSize) {
+          ShuffleDataSegment sds = new ShuffleDataSegment(fileOffset, bufferOffset, bufferSegments);
+          dataFileSegments.add(sds);
+          bufferSegments = Lists.newArrayList();
+          bufferOffset = 0;
+          fileOffset = -1;
+        }
+      } catch (BufferUnderflowException ue) {
+        LOGGER.warn("Read index data under flow", ue);
```
duanmeng, line 190: use `ue.getMessage` to avoid printing the stack?

jerqi: I think `exception stack` will be more helpful when we need to debug.

colinmjj (approved):
LGTM

jerqi:
When we encounter the incomplete index, we just stop reading, not throw an exception.

When we use the storageType MEMORY_HDFS, MEMORY_LOCALFILE_HDFS, after we read data from memory, the data may be flushed to HDFS. So we could read incomplete index data.

No

New ut

PR #79, opened 2022-02-16
[Feature] Enhance the DelegationRssShuffleManager to support hdfs conf file and reset the shuffle manager name

duanmeng:
1. Enhance the candidates checker to support hdfs
2. Reset the shuffle manager name on the fly

It is more feasible to support both local and hdfs and we need to reset shuffle manager name the adaptive ESS when fallback to SortShuffleManager, in which the only the sort shuffle manager is supported.

No

UT & system test

### DIFF
--- a/client-spark/spark2/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
+++ b/client-spark/spark2/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
@@ -71,16 +71,18 @@ public class DelegationRssShuffleManager implements ShuffleManager {
       try {
         shuffleManager = new RssShuffleManager(sparkConf, true);
         sparkConf.set(RssClientConfig.RSS_ENABLED, "true");
+        sparkConf.set("spark.shuffle.manager", RssShuffleManager.class.getCanonicalName());
         LOG.info("Use RssShuffleManager");
         return shuffleManager;
       } catch (Exception exception) {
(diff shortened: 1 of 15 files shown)


coordinator/pom.xml
```diff
@@ -70,6 +70,24 @@
       <scope>test</scope>
       <type>test-jar</type>
     </dependency>
+    <dependency>
+      <groupId>com.tencent.rss</groupId>
+      <artifactId>shuffle-storage</artifactId>
```
colinmjj, line 75: coordinator shouldn't depend on shuffle-storage module

client-spark/spark2/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
```diff
@@ -71,16 +71,18 @@ private ShuffleManager createShuffleManagerInDriver() throws RssException {
       try {
         shuffleManager = new RssShuffleManager(sparkConf, true);
         sparkConf.set(RssClientConfig.RSS_ENABLED, "true");
+        sparkConf.set("spark.shuffle.manager", Constants.RSS_SHUFFLE_MANAGER_NAME);
```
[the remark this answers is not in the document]
duanmeng: sure

coordinator/pom.xml
duanmeng: our hdfs related utils are in this module and  we need to support hdfs candidates file.

colinmjj (approved):
LGTM, depend on UT

PR #95, opened 2022-03-09
[Doc] Add the support information

jerqi:
Add the support information

Provide more communication methods for users.

No

No need


jerqi:
企业微信截图_0edbeb97-b811-4cc5-a458-062443cda60a

colinmjj:
I'm not sure for such information, because I haven't seen it in any apache project.
Slack may be a better place for work together

jerqi:
> I'm not sure for such information, because I haven't seen it in any apache project. Slack may be a better place for work together

Flink Remote Shuffle have similar information. Our users are mainly Chinese.

colinmjj (approved):

jerryshao:
This is not an Apache project, it's OK to leave Tencent specific information and chat group. But we should also suggest user to open issues, which is easy to track.

jerqi:
> This is not an Apache project, it's OK to leave Tencent specific information and chat group. But we should also suggest user to open issues, which is easy to track.
Ok

PR #97, opened 2022-03-14
[Improvement] Remove some integration tests under StorageType.HDFS  mode to …

frankliee:
Remove two integration tests under StorageType.HDFS 

There are some unnecessary UTs that cost time.

No.

All existing UTs.


frankliee:
Do we need to remain  UploadMultiStorage (LOCALFILE_HDFS_2) tests?

frankliee:
Pasted Graphic 2
Pasted Graphic 3

**Reduce about 10 minutes time.**

jerqi:
> Do we need to remain UploadMultiStorage (LOCALFILE_HDFS_2) tests?

We should remove feature and tests at the same time. And if we plan to support streaming application, the feature that server upload shuffle to hdfs useful. But we don't similar plan.

colinmjj:
> Should we remove HDFS StorageType?

@jerqi Yes, storage with HDFS only can't work with task which has small shuffle data segment. And Local_HDFS also can cover HDFS only with proper configuration

colinmjj (approved):

frankliee:
> > Should we remove HDFS StorageType?
> 
> @jerqi Yes, storage with HDFS only can't work with task which has small shuffle data segment. And Local_HDFS also can cover HDFS only with proper configuration

I will add another PR to totally disable HDFS type.

PR #98, opened 2022-03-14
[Bugfix] Exist memory leak when we use aqe

jerqi:
When we finish to use RssShuffleDataIterator, we release the RssShuffleDataIterator's memory.

When we use aqe, if there too many partitions, RssShuffleIterator will occupy much memory, the executor will become out of memory.

No

Manually test. Run 10T's tpcds query1 with `spark.sql.adaptive.advisoryPartitionSizeInBytes 256000000`


colinmjj (dismissed):
LGTM

colinmjj (approved):

PR #99, opened 2022-03-16
[Bugfix] Add timestamp to avoid the conflicts of gc.log after server …

frankliee:
Add a timestamp as the suffix of gc log.

JVM will rewrite gc.log after restart.

No.

By hand.

### DIFF
--- a/bin/start-shuffle-server.sh
+++ b/bin/start-shuffle-server.sh
@@ -83,7 +83,7 @@ JVM_ARGS=" -server \
           -XX:+PrintGCDateStamps \
           -XX:+PrintGCTimeStamps \
           -XX:+PrintGCDetails \
-          -Xloggc:./logs/gc.log"
+          -Xloggc:./logs/gc-%t.log"
 
 ARGS=""
 if [ -f ./conf/log4j.properties ]; then


frankliee:
image

colinmjj (approved):
LGTM

jerqi:
Will we generate too many gc files

frankliee:
> Will we generate too many gc files

This file will only be generated after restart.
There will not exist many files, unless the server crash many times...

jerqi:
> > Will we generate too many gc files
> 
> This file will only be generated after restart. There will not exist many files, unless the server crash many times...

Yes, if we restart too many times.

colinmjj:
> > > Will we generate too many gc files
> > 
> > 
> > This file will only be generated after restart. There will not exist many files, unless the server crash many times...
> 
> Yes, if we restart too many times.

If shuffle server crash, gc log should be kept for investigation.
User should care about too many gc logs which should cause by unhealthy shuffle server

jerqi:
> > > > Will we generate too many gc files
> > > 
> > > 
> > > This file will only be generated after restart. There will not exist many files, unless the server crash many times...
> > 
> > 
> > Yes, if we restart too many times.
> 
> If shuffle server crash, gc log should be kept for investigation. User should care about too many gc logs which should cause by unhealthy shuffle server

I  mean if we should limit the size and the number of gc files.

frankliee:
> gc

This feature is not recommended in JVM ....

https://dzone.com/articles/try-to-avoid-xxusegclogfilerotation

jerqi:
> > gc
> 
> This feature is not recommended in JVM ....
> 
> https://dzone.com/articles/try-to-avoid-xxusegclogfilerotation

Ok, i don't have another suggestions.

jerqi:
may be coordinator also need this parameter.

PR #108, opened 2022-04-14
[Feature] Firestorm supports MapReduce (Part 2 - Map Side)

jerqi:
We add a new MapOutputBuffer to support the use of rss. Actually MapOutputBuffer is a shuffle writer. MapOutputBuffer send data to Shuffle Server.

In real production environments, there are too many MapReduce's jobs. The work is meaningful.

Yes, we bring some new configurations. In another pr, we will update the doc about the use of this feature.

new Uts.


client-mr/src/main/java/org/apache/hadoop/mapred/RssMapOutputCollector.java
```diff
@@ -0,0 +1,203 @@
...
+    keyClass = (Class<K>)jobConf.getMapOutputKeyClass();
+    valClass = (Class<V>)jobConf.getMapOutputValueClass();
+    int sortmb = jobConf.getInt(JobContext.IO_SORT_MB, 100);
+    if ((sortmb & 0x7FF) != sortmb) {
+      throw new IOException(
+          "Invalid \"" + JobContext.IO_SORT_MB + "\": " + sortmb);
+    }
+    partitions = jobConf.getNumReduceTasks();
+    MapTask mapTask = context.getMapTask();
+    long taskAttemptId = (mapTask.getTaskID().getTaskID().getId() << 4) + mapTask.getTaskID().getId();
```
frankliee, line 69: Can you make it easier to understand？
How to convert MR's taskID to RSS's taskID ?

jerqi: I will add some comments.

frankliee: Should we put this conversion in a utility? 
The reader also needs this.

jerqi: > Should we put this conversion in a common utility? The reader also needs this. We need two tool functions.
> 
> 1. Rss taskID -> MR taskID
> 2. MR taskID -> Rss taskID

OK .

client-mr/src/main/java/org/apache/hadoop/mapred/RssMapOutputCollector.java
```diff
@@ -0,0 +1,203 @@
...
+    String storageType = jobConf.get(RssMRConfig.RSS_STORAGE_TYPE);
+    if (StringUtils.isEmpty(storageType)) {
+      throw new RssException("storage type mustn't be empty");
+    }
+
+    Map<Integer, List<ShuffleServerInfo>> partitionToServers = Maps.newHashMap();
+    for (int i = 0; i < partitions; i++) {
+      String servers = jobConf.get(RssMRConfig.RSS_ASSIGNMENT_PREFIX + i);
+      if (StringUtils.isEmpty(servers)) {
+        throw new RssException("assign partition " + i + " is empty");
```
frankliee, line 118: " is empty"  => "should not be empty"

client-mr/src/main/java/org/apache/hadoop/mapred/RssMapOutputCollector.java
```diff
@@ -0,0 +1,203 @@
...
+            serverInfo[0], Integer.parseInt(serverInfo[1]));
+        assignServers.add(sever);
+      }
+      partitionToServers.put(i, assignServers);
+    }
+    SerializationFactory serializationFactory = new SerializationFactory(jobConf);
+    long maxSegmentSize = jobConf.getLong(RssMRConfig.RSS_CLIENT_MAX_SEGMENT_SIZE,
+        RssMRConfig.RSS_CLIENT_DEFAULT_MAX_SEGMENT_SIZE);
+    bufferManager = new SortWriteBufferManager(
+        sortmb << 20,
```
frankliee, line 137: using constant value instead

client-mr/src/main/java/org/apache/hadoop/mapred/RssMapOutputCollector.java
jerqi: Ok

client-mr/src/main/java/org/apache/hadoop/mapred/RssMapOutputCollector.java
jerqi: OK

client-mr/pom.xml
```diff
@@ -63,6 +63,11 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
         </dependency>
+        <dependency>
+            <groupId>net.jpountz.lz4</groupId>
+            <artifactId>lz4</artifactId>
+            <version>1.3.0</version>
```
colinmjj, line 69: dependency should be managed in root pom.xml, and child pom.xml includes it if needed

client-mr/src/main/java/org/apache/hadoop/mapred/MRRssUtils.java
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.hadoop.mapreduce.TaskType;
+
+public class MRRssUtils {
+
+  // Class TaskAttemptId have two field id and mapId, mapId is 4 high byte,
+  // id is low 4 byte.
+  public static long convertTaskAttemptIdToLong(TaskAttemptID taskAttemptID) {
+    long highBytes = ((long)taskAttemptID.getTaskID().getId()) << 32;
+    long lowBytes = taskAttemptID.getId();
+    return highBytes + lowBytes;
```
colinmjj, line 30: Please confirm if highBytes == lowBytes, I'm afraid they are all from ID.getId().

client-mr/src/main/java/org/apache/hadoop/mapred/RssMapOutputCollector.java
```diff
@@ -0,0 +1,204 @@
...
+  private Task.TaskReporter reporter;
+  private Class<K> keyClass;
+  private Class<V> valClass;
+  private Set<Long> successBlockIds = Sets.newConcurrentHashSet();
+  private Set<Long> failedBlockIds = Sets.newConcurrentHashSet();
+  private int partitions;
+  private SortWriteBufferManager bufferManager;
+
+  @Override
+  public void init(Context context) throws IOException, ClassNotFoundException {
```
colinmjj, line 58: too many code in init(), it should be separated in several methods.

client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBufferManager.java
```diff
@@ -0,0 +1,346 @@
...
+    });
+  }
+
+  public void waitSendFinished() {
+    while (!waitSendBuffers.isEmpty()) {
+      sendBuffersToServers();
+    }
+    long start = System.currentTimeMillis();
+    long commitDuration = 0;
+    if (!isMemoryShuffleEnabled) {
```
colinmjj, line 240: I prefer to remove storage types without memory like: local, hdfs, local_hdfs
so, sendCommit() can be avoid

client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBufferManager.java
```diff
@@ -0,0 +1,346 @@
...
+      throw new RssException("record is too big");
+    }
+    buffer.addRecord(key, start, end);
+    memoryUsedSize.addAndGet(length);
+    if (memoryUsedSize.get() > maxMemSize * memoryThreshold) {
+      sendBuffersToServers();
+    }
+    mapOutputRecordCounter.increment(1);
+    mapOutputByteCounter.increment(length);
+    return;
```
colinmjj, line 174: return is unnecessary

client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBuffer.java
```diff
@@ -0,0 +1,196 @@
...
+    long startCopy =  System.currentTimeMillis();
+    sortTime += startCopy - startSort;
+    for (Record<K> record : records) {
+      int beginIndex = (int) (record.getStart() / maxSegmentSize);
+      int beginOffSet = (int) (record.getStart() % maxSegmentSize);
+      int endIndex = (int) (record.getEnd() / maxSegmentSize);
+      int endOffset = (int) (record.getEnd() % maxSegmentSize);
+      if (beginIndex == endIndex) {
+        int length = endOffset - beginOffSet;
+        System.arraycopy(buffers.get(beginIndex).getBuffer(), beginOffSet, data, offset, length);
```
colinmjj, line 73: too many arraycopy, performance is poor

client-mr/pom.xml
jerqi: OK.

client-mr/src/main/java/org/apache/hadoop/mapred/MRRssUtils.java
jerqi: We have ut to confirm this.

colinmjj:
@jerqi I think the problem with current implementation is performance for ordered data in block, and there should be better solution.

client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBuffer.java
jerqi: The mr's arraycopy is less than Spark.

client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBufferManager.java
jerqi: OK

client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBufferManager.java
jerqi: OK, it's another pr.

client-mr/src/main/java/org/apache/hadoop/mapred/RssMapOutputCollector.java
jerqi: OK

client-mr/src/main/java/org/apache/hadoop/mapred/MRRssUtils.java
frankliee: Is 32 bits too large for attempt number ?

client-mr/src/main/java/org/apache/hadoop/mapred/RssMapOutputCollector.java
```diff
@@ -0,0 +1,204 @@
...
+          throw new RssException("partition " + i + " server info isn't right");
+        }
+        ShuffleServerInfo sever = new ShuffleServerInfo(StringUtils.join(serverInfo, "-"),
+            serverInfo[0], Integer.parseInt(serverInfo[1]));
+        assignServers.add(sever);
+      }
+      partitionToServers.put(i, assignServers);
+    }
+    SerializationFactory serializationFactory = new SerializationFactory(jobConf);
+    long maxSegmentSize = jobConf.getLong(RssMRConfig.RSS_CLIENT_MAX_SEGMENT_SIZE,
```
colinmjj, line 135: maxSegmentSize = max(size of current record, size in config)

client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBuffer.java
```diff
@@ -0,0 +1,196 @@
...
+import org.apache.hadoop.io.RawComparator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SortWriteBuffer<K, V> extends OutputStream  {
+
+  private static final Logger LOG = LoggerFactory.getLogger(SortWriteBuffer.class);
+  private long copyTime = 0;
+  private final List<WrappedBuffer> buffers = Lists.newArrayList();
+  private final List<Record<K>> records = Lists.newArrayList();
```
colinmjj, line 36: one record will be stored in both buffers(for serialized) and records(for order)?

client-mr/src/main/java/org/apache/hadoop/mapred/RssMapOutputCollector.java
jerqi: This is init function, we don't know the size of current record.

client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBuffer.java
jerqi: No, we only store keys in the records.

client-mr/src/main/java/org/apache/hadoop/mapred/RssMapOutputCollector.java
colinmjj: when create wrappedBuffer size = max(size of current record, size in config)

client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBufferManager.java
```diff
@@ -0,0 +1,352 @@
...
+    this.maxMemSize = maxMemSize;
+    this.taskAttemptId = taskAttemptId;
+    this.batch = batch;
+    this.keySerializer = keySerializer;
+    this.valSerializer = valSerializer;
+    this.comparator = comparator;
+    this.memoryThreshold = memoryThreshold;
+    this.appId = appId;
+    this.shuffleWriteClient = shuffleWriteClient;
+    this.sendCheckInterval = sendCheckInterval;
```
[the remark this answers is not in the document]
jerqi: No, but this parameter is necessary.

client-mr/src/main/java/org/apache/hadoop/mapred/RssMapOutputCollector.java
jerqi: What's the advantages?

jerqi: I use fix length buffer to avoid memory fragments.

colinmjj: here is case in spark, the record size > size in config, and the data can't be stored.
job failed.

jerqi: > here is case in spark, the record size > size in config, and the data can't be stored. job failed.

In mr, the record will be split to multiple buffers.

client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBuffer.java
colinmjj: Data size in records is ignored, and it will cause memory problem.
For example, in "word count" case, data size of key can't be ignored.

jerqi: > Data size in records is ignored, and it will cause memory problem. For example, in "word count" case, data size of key can't be ignored.
Let me think of another method.

jerqi: > > Data size in records is ignored, and it will cause memory problem. For example, in "word count" case, data size of key can't be ignored.
> > 
Let me think of another method.

jerqi: It seems two methods to solve this,  We don't serialize key at fist, when we sort finish, we serialize the keys.But it will be too difficult.I think it will be more easy if we sort the block  in the reduce side.

colinmjj: > 

If sort is happened only in reduce task, can't reuse sort implementation from MR.
For this problem, we need get a better solution first.

client-mr/src/main/java/org/apache/hadoop/mapred/MRRssUtils.java
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * https://opensource.org/licenses/Apache-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OF ANY KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package org.apache.hadoop.mapred;
```
frankliee, line 19: It is better put MRRssUtils in org.apache.hadoop.mapreduce

frankliee: And JobID should use org.apache.hadoop.mapreduce.JobID

jerqi: Why?

frankliee: TaskAttemptID.getJob() returns org.apache.hadoop.mapreduce.JobID

jerqi: OK

client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBuffer.java
jerqi: modify as we discussed.

jerqi:
@colinmjj @frankliee All comments are addressed.

client-mr/src/main/java/org/apache/hadoop/mapreduce/MRRssUtils.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.hadoop.mapred.JobConf;
+
+import com.tencent.rss.client.api.ShuffleWriteClient;
+import com.tencent.rss.client.factory.ShuffleClientFactory;
+
+public class MRRssUtils {
+
+  // Class TaskAttemptId have two field id and mapId, mapId is 4 high byte,
+  // id is low 4 byte.
+  public static long convertTaskAttemptIdToLong(TaskAttemptID taskAttemptID) {
```
colinmjj, line 30: taskId will be stored in bitmap at shuffle server, it should be designed to reduce memory cost. 
with mapId as high byte, I think it will waste more memory

jerqi: OK

colinmjj (dismissed):
LGTM

colinmjj (approved):

PR #115, opened 2022-04-22
[Feature] Firestorm supports MapReduce (Part 4 -  Integration Tests)

jerqi:
We add some integration tests for mr, such as word count , secondary sort. And we fix some problems, such heartbeat, map side data format, etc.

We need integration tests to ensure the correctness of our system.

No

Just test


colinmjj (approved):
LGTM, still need test with TeraSort & TPC-DS to validate performance and correctness.

PR #117, opened 2022-04-24
[Bug] Fix NPE problems occasionally

jerqi:
Sometimes bufferManager will be removed in another thread, so it will throw npe.

Avoid npe problems.

No

No, just add some checkers.


colinmjj (dismissed):
LGTM

frankliee (approved):

PR #118, opened 2022-04-24
[Improvement] Remove useless StorageTypes

jerqi:
Remove useless Storage, such as LOCALFILE, HDFS, LOCALFILE_HDFS, LOCALFILE_HDFS2

These StorageTypes don't have enough good performance, and we need many time to test them

Yes, users can't use these StorageTypes which are removed.

GA passed.


colinmjj:
Too many fix are included, eg, remove storage type, remove commit action, remove multiple storage feature, etc.
For LOCALFILE, HDFS, LOCALFILE_HDFS, they can be useful in test case.
It's better to do the source code cleanup step by step.

jerqi:
> Too many fix are included, eg, remove storage type, remove commit action, remove multiple storage feature, etc. For LOCALFILE, HDFS, LOCALFILE_HDFS, they can be useful in test case. It's better to do the source code cleanup step by step.

For Tests, we use MEMORY_LOCALFILE and MEMORY_HDFS instead.

PR #125, opened 2022-05-06
[Improvement] Optimize quorum write by skipping unnecessary blocks

frankliee:
Write client send blocks in two rounds.
In the primary round, we only send [0, replicaWrite) data.
In the secondary round, we send [replicaWrite, replica) data.
When  the primary round is totally successful, the secondary round can be skipped. 

Without this patch, the  quorum writer will send whole replica times of blocks even all RPCs are successful.

No.

All previous UTs and new UTs.

### DIFF
--- a/client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRConfig.java
+++ b/client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRConfig.java
@@ -50,6 +50,10 @@ public class RssMRConfig {
       MR_RSS_CONFIG_PREFIX + RssClientConfig.RSS_DATA_REPLICA_READ;
   public static final int RSS_DATA_REPLICA_READ_DEFAULT_VALUE =
       RssClientConfig.RSS_DATA_REPLICA_READ_DEFAULT_VALUE;
+  public static final String RSS_DATA_REPLICA_SKIP =
+      MR_RSS_CONFIG_PREFIX + RssClientConfig.RSS_DATA_REPLICA_SKIP;
+  public static boolean RSS_DATA_REPLICA_SKIP_DEFAULT_VALUE =
+      RssClientConfig.RSS_DATA_REPLICA_SKIP_DEFAULT_VALUE;
   public static final String RSS_HEARTBEAT_INTERVAL =
       MR_RSS_CONFIG_PREFIX + RssClientConfig.RSS_HEARTBEAT_INTERVAL;
   public static final long RSS_HEARTBEAT_INTERVAL_DEFAULT_VALUE =
--- a/client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRConfig.java
+++ b/client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRConfig.java
@@ -50,10 +50,10 @@ public class RssMRConfig {
       MR_RSS_CONFIG_PREFIX + RssClientConfig.RSS_DATA_REPLICA_READ;
   public static final int RSS_DATA_REPLICA_READ_DEFAULT_VALUE =
       RssClientConfig.RSS_DATA_REPLICA_READ_DEFAULT_VALUE;
-  public static final String RSS_DATA_REPLICA_SKIP =
-      MR_RSS_CONFIG_PREFIX + RssClientConfig.RSS_DATA_REPLICA_SKIP;
-  public static boolean RSS_DATA_REPLICA_SKIP_DEFAULT_VALUE =
-      RssClientConfig.RSS_DATA_REPLICA_SKIP_DEFAULT_VALUE;
+  public static final String RSS_DATA_REPLICA_SKIP_ENABLED =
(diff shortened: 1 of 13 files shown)


colinmjj (approved):
LGTM

jerqi:
Do you add the function switch? And other replicas case.

frankliee:
> Do you add the function switch? And other replicas case.

More cases are adding. 

But is the switch really necessary? Nobody wants to send more data...

frankliee:
> > Do you add the function switch? And other replicas case.
> 
> More cases are adding.
> 
> But is the switch really necessary? Nobody wants to send more data...

And this interface is ready complicated.
ShuffleWriteClientImpl(String clientType, int retryMax, long retryIntervalMax, int heartBeatThreadNum, int replica, int replicaWrite, int replicaRead)

jerqi:
If one server crash, origin implement have better than this.

frankliee:
> If one server crash, origin implement have better than this.

ok

frankliee:
> If one server crash, origin implement have better than this.

This new option will be added after the unified config PR has been merged.

jerqi:
> > If one server crash, origin implement have better than this.
> 
> This new option will be added after the unified config PR has been merged.

OK

jerqi:
LGTM+2

jerqi (dismissed):

client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRConfig.java
```diff
@@ -50,6 +50,10 @@ public class RssMRConfig {
       MR_RSS_CONFIG_PREFIX + RssClientConfig.RSS_DATA_REPLICA_READ;
   public static final int RSS_DATA_REPLICA_READ_DEFAULT_VALUE =
       RssClientConfig.RSS_DATA_REPLICA_READ_DEFAULT_VALUE;
+  public static final String RSS_DATA_REPLICA_SKIP =
```
jerqi, line 53: RSS_DATA_REPLICA_SKIP_ENABLED?

client/src/main/java/com/tencent/rss/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -156,41 +172,72 @@ public SendShuffleDataResult sendShuffleData(String appId, List<ShuffleBlockInfo
...
     // todo: better to have fallback solution when send to multiple servers
-    sendShuffleDataAsync(appId, serverToBlocks, serverToBlockIds, successBlockIds, failedBlockIds);
+
+    // sent the primary round of blocks.
+    boolean isAllSuccess = sendShuffleDataAsync(
+        appId, primaryServerToBlocks, primaryServerToBlockIds, blockIdsTracker);
+
+    // The secondary round of blocks is sent only when the primary group issues failed sending.
+    // This should be infrequent.
+    if (!isAllSuccess && replicaSkip) {
```
jerqi, line 225: `replicaSkip` is not necessary.

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -78,6 +78,7 @@ public class RssShuffleManager implements ShuffleManager {
   private final int dataReplica;
   private final int dataReplicaWrite;
   private final int dataReplicaRead;
+  private final boolean dataReplicaSkip;
```
jerqi, line 81: enabled.

client/src/main/java/com/tencent/rss/client/impl/ShuffleWriteClientImpl.java
frankliee: This is the new option

jerqi: If replicaSkip is enable, collection is empty.We won't send any rpc.It's up to you.

frankliee: Ok, I will use `!primaryServerToBlocks.isEmpty()` instead.

client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRConfig.java
frankliee: OK

client-spark/spark3/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
frankliee: OK

client/src/main/java/com/tencent/rss/client/impl/ShuffleWriteClientImpl.java
jerqi: SendaryServerToBlocks.isEmpty()?

frankliee: Yes, copy error...

jerqi (dismissed):

jerqi (approved):

PR #126, opened 2022-05-06
[Improvement] Unify config name for RSS client

colinmjj:
MR and Spark are both supported now. The config name should be unified for better maintenance.
The config name will be defined as engine prefix + rss config

For better maintenance

No, but rss.bath.path is changed to rss.remote.storage.path

new UTs added


client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
```diff
@@ -0,0 +1,128 @@
...
+
+import java.util.Set;
+
+import com.google.common.collect.Sets;
+
+import com.tencent.rss.client.util.RssClientConfig;
+
+public class RssSparkConfig {
+
+  public static String SPARK_RSS_CONFIG_PREFIX = "spark.";
```
duanmeng, line 29: `public static final`？

client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRConfig.java
```diff
@@ -18,51 +18,90 @@
...
-  public static final int RSS_INDEX_READ_LIMIT_DEFAULT_VALUE = 500;
-  public static String RSS_CLIENT_READ_BUFFER_SIZE = "mapreduce.rss.client.read.buffer.size";
   // When the size of read buffer reaches the half of JVM region (i.e., 32m),
   // it will incur humongous allocation, so we set it to 14m.
-  public static String RSS_CLIENT_READ_BUFFER_SIZE_DEFAULT_VALUE = "14m";
+  public static String RSS_CLIENT_READ_BUFFER_SIZE_DEFAULT_VALUE =
+      RssClientConfig.RSS_CLIENT_READ_BUFFER_SIZE_DEFAULT_VALUE;
 
-  public static String RSS_DYNAMIC_CLIENT_CONF_ENABLED = "mapreduce.rss.dynamicClientConf.enabled";
+  public static String RSS_DYNAMIC_CLIENT_CONF_ENABLED =
```
frankliee, line 99: Do we need to put "rss.dynamicClientConf.enabled" in RssClientConfig?

client-spark/common/src/main/java/org/apache/spark/shuffle/RssSparkConfig.java
colinmjj: ok

client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRConfig.java
colinmjj: ok

frankliee (approved):

PR #137, opened 2022-05-13
[Bug] Fix NPE problem in coordinator if there has no remote storage

colinmjj:
There has NPE problem if there has no remote storage

Bug fix

No

New UT is added

### DIFF
--- a/coordinator/src/main/java/com/tencent/rss/coordinator/ApplicationManager.java
+++ b/coordinator/src/main/java/com/tencent/rss/coordinator/ApplicationManager.java
@@ -52,6 +52,8 @@ public class ApplicationManager {
   private Map<String, String> remoteStorageToHost = Maps.newConcurrentMap();
   private Set<String> availableRemoteStoragePath = Sets.newConcurrentHashSet();
   private ScheduledExecutorService scheduledExecutorService;
+  // it's only for test case to check if status check has problem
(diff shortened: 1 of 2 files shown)


coordinator/src/main/java/com/tencent/rss/coordinator/ApplicationManager.java
```diff
@@ -161,21 +163,23 @@ protected synchronized void incRemoteStorageCounter(String remoteStoragePath) {
...
+        if (count < 0) {
+          LOG.warn("Unexpected counter for remote storage: %s, which is %i, reset to 0",
+              storagePath, count);
+          atomic.set(0);
+        }
+      } else {
+        LOG.warn("Can't find counter for remote storage: {}", storagePath);
+        remoteStoragePathCounter.putIfAbsent(storagePath, new AtomicInteger(0));
+      }
+      if (remoteStoragePathCounter.get(storagePath).get() == 0
```
jerqi, line 179: Is it possible that remoteStoragePathCounter.get  return null?

colinmjj: if null, there will be `remoteStoragePathCounter.putIfAbsent()` to add new object before this statement.

jerqi (approved):

PR #138, opened 2022-05-13
[Improvement]  Filter the interface address that don't have broadcast address

jerqi:
Filter the interface address that don't have broadcast address

Some machines have multiple interface address, but some interface address don't have broadcast address, we can't use it to communicate, so we should filter it.

No

Manual Test


frankliee:
Do we need a warning or info log for this situation？

jerqi:
> Do we need a warning or info log for this situation？

I don't think we need. Because it's not warning.

colinmjj:
It's better to add log to show which ip is parsed and which one is used.

jerqi:
> It's better to add log to show which ip is parsed and which one is used.
OK.

colinmjj (approved):
LGTM

PR #139, opened 2022-05-14
[Bugfix] Shade jar should contain lz4 jar

jerqi:
Include lz4 jar in our shade jar

If we don't have this patch, we will encounter exception that we can't load Lz4 class.

No

Manual test


frankliee:
LGTM

frankliee (approved):

PR #140, opened 2022-05-14
[Bugfix] AppMaster configuration doesn't work

jerqi:
After we modify the configuration, we should update the hdfs file and the local file. And we don't support recovery now, so we close it.

If we don't have this patch,  we cannot use new configuration.

No

Manual test and new UT

### DIFF
--- a/client-mr/src/main/java/org/apache/hadoop/mapreduce/v2/app/RssMRAppMaster.java
+++ b/client-mr/src/main/java/org/apache/hadoop/mapreduce/v2/app/RssMRAppMaster.java
@@ -18,6 +18,10 @@
 
 package org.apache.hadoop.mapreduce.v2.app;
 
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


client-mr/src/main/java/org/apache/hadoop/mapreduce/v2/app/RssMRAppMaster.java
```diff
@@ -160,16 +163,24 @@ public Thread newThread(Runnable r) {
...
       throw new RuntimeException("jobDir is empty");
     }
     Path jobConfFile = new Path(jobDirStr, MRJobConfig.JOB_CONF_FILE);
+    updateConf(conf, jobConfFile);
+    // remove org.apache.hadoop.mapreduce.v2.app.MRAppMaster
+    ArrayUtils.remove(args, 0);
+    MRAppMaster.main(args);
+  }
+
+  static void updateConf(JobConf conf, Path jobConfFile) {
```
frankliee, line 183: We need a comment about it.

jerqi: OK

frankliee (approved):

PR #141, opened 2022-05-16
[Bugfix] If there is no shuffle, we shouldn't acess rss

jerqi:
If there is no shuffle, we shouldn't acess rss, otherwise we will encounter an error

If a application has no shuffle, it will occur a error

No

Manual test

### DIFF
--- a/client-mr/src/main/java/org/apache/hadoop/mapreduce/v2/app/RssMRAppMaster.java
+++ b/client-mr/src/main/java/org/apache/hadoop/mapreduce/v2/app/RssMRAppMaster.java
@@ -63,103 +63,105 @@ public class RssMRAppMaster {
 
     JobConf conf = new JobConf(new Path(MRJobConfig.JOB_CONF_FILE));
     int numReduceTasks = conf.getInt(MRJobConfig.NUM_REDUCES, 0);
-    String coordinators = conf.get(RssMRConfig.RSS_COORDINATOR_QUORUM);
+    if (numReduceTasks > 0) {
+      String coordinators = conf.get(RssMRConfig.RSS_COORDINATOR_QUORUM);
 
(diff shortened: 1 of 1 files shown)


client-mr/src/main/java/org/apache/hadoop/mapreduce/v2/app/RssMRAppMaster.java
```diff
@@ -60,112 +60,114 @@ public static void main(String[] args) {
...
-        1, Sets.newHashSet(Constants.SHUFFLE_SERVER_VERSION));
-    Map<ShuffleServerInfo, List<PartitionRange>> serverToPartitionRanges = response.getServerToPartitionRanges();
-    final ScheduledExecutorService scheduledExecutorService = Executors.newSingleThreadScheduledExecutor(
-        new ThreadFactory() {
-          @Override
-          public Thread newThread(Runnable r) {
-            Thread t = Executors.defaultThreadFactory().newThread(r);
-            t.setDaemon(true);
-            return t;
+    if (numReduceTasks > 0) {
```
frankliee, line 63: What  about numReduceTasks == 0 ???

jerqi: `numReduceTasks == 0` means no shuffle

frankliee: Which MR applications have no shuffle?

jerqi: map only, distcp

frankliee (approved):

PR #142, opened 2022-05-17
[Bug] Update hostname to valid metric name

colinmjj:
As metric name, host name has invalid character like `-`, `.`
Replace them with `_` to generate valid metric name

Bug fix

No

UT added

### DIFF
--- a/common/src/main/java/com/tencent/rss/common/util/RssUtils.java
+++ b/common/src/main/java/com/tencent/rss/common/util/RssUtils.java
@@ -267,4 +267,14 @@ public class RssUtils {
       throw new RuntimeException("Replica config is unsafe, recommend replica.write + replica.read > replica");
     }
   }
+
+  // the method is used to transfer hostname to metric name.
+  // With standard of hostname, `-` and `.` will be included in hostname,
+  // but they are not valid for metric name, replace them with '_'
+  public static String getMetricNameForHostName(String hostName) {
+    if (hostName == null) {
+      return "";
+    }
+    return hostName.replaceAll("[\\.-]", "_");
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/tencent/rss/common/util/RssUtils.java
```diff
@@ -267,4 +267,14 @@ public static void checkQuorumSetting(int replica, int replicaWrite, int replica
...
   }
+
+  // the method is used to transfer hostname to metric name.
+  // With standard of hostname, `-` and `.` will be included in hostname,
+  // but they are not valid for metric name, replace them with '_'
+  public static String getMetricNameForHostName(String hostName) {
+    if (hostName == null) {
+      return "";
+    }
+    return hostName.replaceAll("[\\.-]", "_");
```
frankliee, line 278: What about ' '(i.e., blank)?

colinmjj: For hostname, standards(eg, RFC 1035) which only contains `[a-z], -, .` are applied, the other character is not allowed. If the rule is broken, update the method according to new rule.

frankliee: That is OK.
LT2M.

frankliee (approved):

colinmjj:
What changes were proposed in this pull request?
As metric name, host name has invalid character like -, .
Replace them with _ to generate valid metric name

Why are the changes needed?
Bug fix

Does this PR introduce any user-facing change?
No

How was this patch tested?
UT added

PR #151, opened 2022-05-24
[Bugfix] Fix taskAttemptId length error

jerqi:
Rss TaskAttemptId length is 20, we should guarantee the map task id length is less than that value. And we adjust the length of task to run more tasks.

If we don't have this patch,  task cannot retry succesfully. Task will throw exception
Can't support taskAttemptId[4294976976], the max value should be 1048575
        at com.tencent.rss.client.util.ClientUtils.getBlockId(ClientUtils.java:43)
        at org.apache.hadoop.mapred.SortWriteBufferManager.createShuffleBlock(SortWriteBufferManager.java:292)

No

Add new UT

### DIFF
--- a/client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRUtils.java
+++ b/client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRUtils.java
@@ -38,19 +38,29 @@ import com.tencent.rss.common.exception.RssException;
 public class RssMRUtils {
 
   private static final Logger LOG = LoggerFactory.getLogger(RssMRUtils.class);
+  private static int MAX_TASK_LENGTH = 17;
+  private static int MAX_ATTEMPT_LENGTH = 3;
+  private static long MAX_TASK_ID = (1 << MAX_TASK_LENGTH) - 1;
+  private static long MAX_ATTEMPT_ID = ( 1 << MAX_ATTEMPT_LENGTH) - 1;
 
-  // Class TaskAttemptId have two field id and mapId, mapId is 4 low byte,
-  // id is high 4 byte.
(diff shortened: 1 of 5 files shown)


client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRUtils.java
```diff
@@ -38,19 +38,29 @@
...
   private static final Logger LOG = LoggerFactory.getLogger(RssMRUtils.class);
+  private static int MAX_TASK_LENGTH = 19;
+  private static int MAX_ATTEMPT_LENGTH = 2;
+  private static long MAX_TASK_ID = (1 << MAX_TASK_LENGTH) - 1;
+  private static long MAX_ATTEMPT_ID = (1 << MAX_ATTEMPT_LENGTH) - 1;
 
-  // Class TaskAttemptId have two field id and mapId, mapId is 4 low byte,
-  // id is high 4 byte.
+  // Class TaskAttemptId have two field id and mapId, rss taskAttemptID have 21 bits,
+  // mapId is 19 bits, id is 2 bits.
```
frankliee, line 47: Is 2-bit for attemptID is too tricky ?

jerqi: When we increate a bit, we will increate double meta memory.

jerqi:
@colinmjj If you have another suggestion, I'll continue to test this server.

colinmjj (approved):
LGTM, I think there should be another PR to trace the memory of metadata.

jerqi:
> LGTM, I think there should be another PR to trace the memory of metadata.

OK

PR #153, opened 2022-05-25
[Minor] Fix integer division in floating-point context

kaijchen:

Fix integer division in floating-point context.

Fix precision loss and potential overflow.

No.

https://github.com/kaijchen/Firestorm/actions/runs/2384201966


jerqi (approved):
Thanks for your contribution, but we will remove these code about upload in the future

kaijchen:
Thanks @jerqi for the review.

PR #155, opened 2022-05-25
[Infra] Configure checkstyle in pom.xml

kaijchen:

Properly configure checkstyle plugin in pom.xml.

For better enforcing code style check.

No.

Before:
https://github.com/Tencent/Firestorm/runs/6591934064?check_suite_focus=true#step:4:687

After this change, much more violations were found:
https://github.com/kaijchen/Firestorm/runs/6593111997?check_suite_focus=true#step:4:2640

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -773,7 +773,7 @@
           <suppressionsLocation>checkstyle-suppressions.xml</suppressionsLocation>
           <encoding>UTF-8</encoding>
           <consoleOutput>true</consoleOutput>
-          <failsOnError>false</failsOnError>
+          <failsOnError>true</failsOnError>
           <linkXRef>false</linkXRef>
           <includeTestSourceDirectory>true</includeTestSourceDirectory>
         </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -769,19 +769,13 @@
         <artifactId>maven-checkstyle-plugin</artifactId>
         <version>${maven-checkstyle-plugin.version}</version>
         <configuration>
-          <sourceDirectories>
-            <directory>${basedir}/src/main/java</directory>
-          </sourceDirectories>
-          <testSourceDirectories>
-            <directory>${basedir}/src/test/java</directory>
-          </testSourceDirectories>
           <configLocation>checkstyle.xml</configLocation>
+          <suppressionsLocation>checkstyle-suppressions.xml</suppressionsLocation>
           <encoding>UTF-8</encoding>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -769,19 +769,13 @@
...
-          </sourceDirectories>
-          <testSourceDirectories>
-            <directory>${basedir}/src/test/java</directory>
-          </testSourceDirectories>
           <configLocation>checkstyle.xml</configLocation>
+          <suppressionsLocation>checkstyle-suppressions.xml</suppressionsLocation>
           <encoding>UTF-8</encoding>
           <consoleOutput>true</consoleOutput>
-          <failsOnError>true</failsOnError>
+          <failsOnError>false</failsOnError>
```
jerqi, line 776: why do we need `false` option?

kaijchen: This option does not matter, because there is no `<execution>` config in checkstyle.
If it has `<execution>` config, it will fail `mvn package` if checkstyle fails if set `true`.

jerqi: In origin configuration, we seems not to check test file import order. But if we have wrong import order in our main code, checkstyle will fail. In new configuration, we can check import order in test code, but checkstyle didn't fail. Is there anything wrong?

kaijchen: > In origin configuration, we seems not to check test file import order. But if we have wrong import order in our main code, checkstyle will fail. In new configuration, we can check import order in test code, but checkstyle didn't fail. Is there anything wrong?

Can you give an example link to CI (before and after)?

jerqi: We can see your link
企业微信截图_409b5f2d-3e53-490b-b59e-81a3fc3afdf6
Although there some warning, but checkstyle didn't fail.

kaijchen: This is because of checkstyle.xml

https://github.com/Tencent/Firestorm/blob/795f75a7030a4ac204aaef20f0f70caba1a39d0c/checkstyle.xml#L26

kaijchen: We can turn it on while we are fixing checkstyle violations.

jerqi: How can I turn it on?

kaijchen: > How can I turn it on?

Remove
https://github.com/Tencent/Firestorm/blob/795f75a7030a4ac204aaef20f0f70caba1a39d0c/checkstyle.xml#L26

kaijchen, line 776:
```suggestion
          <failsOnError>true</failsOnError>
```

PR #160, opened 2022-05-30
[Bugfix] executor should get remote storage from ShuffleHandle

duanmeng:
Executor get remote storage info from RssShuffleHadle.

The driver's ShuffleManager register shuffle and fetch the remote storage and put then in the RssShuffleHandle. The remote storage in executor's ShuffleManager is useless and the task should get remote storage from RssShuffleHandle when it get reader.

No

Normal  ut and system test


client-spark/spark2/src/main/java/org/apache/spark/shuffle/RssShuffleManager.java
```diff
@@ -338,9 +337,15 @@ public <K, C> ShuffleReader<K, C> getReader(ShuffleHandle handle,
           + blockIdBitmap.getLongCardinality() + " blockIds for shuffleId[" + shuffleId + "], partitionId["
           + startPartition + "]");
 
+      final RemoteStorageInfo shuffleRemoteStorageInfo = rssShuffleHandle.getRemoteStorage();
```
jerqi, line 340: Could we extract a function in module `spark common`

duanmeng: ok

colinmjj (approved):
LGTM

PR #161, opened 2022-05-30
[Bugfix] Map side avoid producing too many blocks

jerqi:
We change the sort order, we should send bigger size block, otherwise we will produce too many blocks. And it will take too much time to send them.

If we don't have this patch,  the task will fail because of timeout.

No

New ut


colinmjj (dismissed):
LGTM

frankliee (approved):

PR #162, opened 2022-05-31
[Bugfix] MR Reduce should only fetch one attempt of each map task

frankliee:
MR Reduce should only fetch one attempt of each map task.

Without this PR, the Reduce will assume to accept all attempt of each map task and fail in taskID check.

No.

By hand.

### DIFF
--- a/client-mr/src/main/java/org/apache/hadoop/mapreduce/task/reduce/RssEventFetcher.java
+++ b/client-mr/src/main/java/org/apache/hadoop/mapreduce/task/reduce/RssEventFetcher.java
@@ -19,8 +19,8 @@
 package org.apache.hadoop.mapreduce.task.reduce;
 
 import java.io.IOException;
-import java.util.HashSet;
-import java.util.Set;
+import java.util.LinkedList;
+import java.util.List;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -43,8 +43,8 @@ public class RssEventFetcher<K,V> {
   private final int maxEventsToFetch;
   private JobConf jobConf;
 
-  private Set<TaskAttemptID> successMaps = new HashSet<TaskAttemptID>();
-  private Set<TaskAttemptID> obsoleteMaps = new HashSet<TaskAttemptID>();
+  private List<TaskAttemptID> successMaps = new LinkedList<>();
+  private List<TaskAttemptID> obsoleteMaps = new LinkedList<>();
   private int tipFailedCount = 0;
   private final int totalMapsCount;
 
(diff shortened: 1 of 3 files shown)


client-mr/src/main/java/org/apache/hadoop/mapreduce/task/reduce/RssEventFetcher.java
```diff
@@ -73,18 +73,21 @@ public Roaring64NavigableMap fetchAllRssTaskIds() {
...
     for (TaskAttemptID taskAttemptID: successMaps) {
       if (!obsoleteMaps.contains(taskAttemptID)) {
-        long rssTaskId = RssMRUtils.convertTaskAttemptIdToLong(taskAttemptID);
-        taskIdBitmap.addLong(rssTaskId);
         int mapIndex = taskAttemptID.getTaskID().getId();
-        if (mapIndex < totalMapsCount) {
-          mapIndexBitmap.addLong(mapIndex);
-        } else {
-          throw new IllegalStateException(errMsg);
+        // we only need to just one attempt of each map task.
```
jerqi, line 77: one successful attempt?

frankliee: This logic is to skip redundancy successful attempt.

Otherwise   
if (taskIdBitmap.getLongCardinality() + tipFailedCount != totalMapsCount) {
      throw new IllegalStateException(errMsg);
}

jerqi: I mean you should modify the comment.

frankliee: Updated

client-mr/src/main/java/org/apache/hadoop/mapreduce/task/reduce/RssEventFetcher.java
```diff
@@ -73,18 +73,21 @@ public Roaring64NavigableMap fetchAllRssTaskIds() {
     String errMsg = "TaskAttemptIDs are inconsistent with map tasks";
     for (TaskAttemptID taskAttemptID: successMaps) {
       if (!obsoleteMaps.contains(taskAttemptID)) {
-        long rssTaskId = RssMRUtils.convertTaskAttemptIdToLong(taskAttemptID);
-        taskIdBitmap.addLong(rssTaskId);
         int mapIndex = taskAttemptID.getTaskID().getId();
```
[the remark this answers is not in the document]
frankliee: Ok, you can test it on the cluster first.

PR #164, opened 2022-06-06
[Improvement] Optimize taskAttemptId format in MR application

jerqi:
We use high bits to express task attempt times, task attempt times don't vary frequently. So we should not increase memory of bitmap.

In our production environment, task often will be preempted. Four times is not enough, so we increase max attempts to  64.

No

Exist UT and manual test


colinmjj (approved):
LGTM

PR #163, opened 2022-06-01
[Improvement] Supports MR application  re-execute

jerqi:
### What changes were proposed in this pull request?
We increase a extra configuration to store rss configuration information, we don't modify the job.xml on the HDFS any more.

### Why are the changes needed?
If we don't have this patch, mr application cannot re-execute successfully. We will encounter the exception

### Does this PR introduce _any_ user-facing change?
No

### How was this patch tested?
new UT

PR #165, opened 2022-06-06
[Bugfix] Fix MR configuration problems

jerqi:
We don't use RssConfig to override JobConf, Because some configuration will be overrided by default configuration. User configuration don't take effect.

#163 introduced this problem. If we don't have this patch, some configuration will not take effect, like `io.sort.mb`.

No

Manual test

### DIFF
--- a/client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRUtils.java
+++ b/client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRUtils.java
@@ -145,4 +145,24 @@ public class RssMRUtils {
       }
     }
   }
+
+  public static int getInt(JobConf priorityConf, JobConf normalConf, String key, int defaultValue) {
+    return priorityConf.getInt(key, normalConf.getInt(key, defaultValue));
+  }
+
+  public static long getLong(JobConf priorityConf, JobConf normalConf, String key, long defaultValue) {
+    return priorityConf.getLong(key, normalConf.getLong(key, defaultValue));
+  }
+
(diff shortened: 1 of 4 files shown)


client-mr/src/main/java/org/apache/hadoop/mapreduce/RssMRUtils.java
```diff
@@ -145,4 +145,24 @@ public static void applyDynamicClientConf(JobConf jobConf, Map<String, String> c
       }
     }
   }
+
+  public static int getInt(JobConf priorityConf, JobConf normalConf, String key, int defaultValue) {
```
frankliee, line 149: The name of priorityConf and normalConf are strange.

jerqi: `FirstConf SecondConf` may be better?

frankliee: conf and defaultConf?

jerqi: Not good. `rssJobConf and jobConf` ?

frankliee: rssJobConf and mrJobConf ？

jerqi: ok.

frankliee:
LT2M

frankliee (approved):

PR #167, opened 2022-06-08
[Bugfix] In MR, OBSOLETE/FAILED/KILLED event will not revoke success attempts

frankliee:
In MR, OBSOLETE/FAILED/KILLED event will not revoke success attempts.

Different from default MR shuffle, in rss shuffle, OBSOLETE/FAILED/KILLED events, 
we should not revoke successful attempts (maybe caused by "bad node" after map success).
This is because these map tasks have already sent their data to rss server.

 No

Update Previous UT

### DIFF
--- a/client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBufferManager.java
+++ b/client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBufferManager.java
@@ -166,7 +166,7 @@ public class SortWriteBufferManager<K, V> {
     if (length > maxMemSize) {
       throw new RssException("record is too big");
     }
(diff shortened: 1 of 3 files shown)


colinmjj:
Is it possible to add a test case for such situation?

frankliee:
> Is it possible to add a test case for such situation?

EvenFetcherTest's obsoletedAndTipFailedEventFetch() has simulated this situation.
succ tasks [0, 100) 
obsoleted {70, 71}
tipFailed {89}

The reduce should still read {70, 71}.

colinmjj (dismissed):
LGTM

jerqi:
Do we cllose this pr?

PR #168, opened 2022-06-09
[Bugfix] Fix mapTask oom problems

jerqi:
We do two things:
1. increase partition level memory limit, If partition buffer memory exceed this limit, we will send this buffer.
2. decrease temporary memory, when we don't need the buffer, we will release this buffer immediately.

If we don't have this patch, we will encounter oom problems

Yes, we increase two configuration
mapreduce.rss.writer.buffer.size means partition level memory limit,
mapreduce.rss.client.memory.threshold means that we can use memory percentage of total sort memory.

New UT

### DIFF
--- a/client-mr/src/main/java/org/apache/hadoop/mapred/RssMapOutputCollector.java
+++ b/client-mr/src/main/java/org/apache/hadoop/mapred/RssMapOutputCollector.java
@@ -68,6 +68,13 @@ public class RssMapOutputCollector<K extends Object, V extends Object>
     partitions = mrJobConf.getNumReduceTasks();
     MapTask mapTask = context.getMapTask();
     JobConf rssJobConf = new JobConf(RssMRConfig.RSS_CONF_FILE);
+    double sortThreshold = RssMRUtils.getDouble(rssJobConf, mrJobConf, RssMRConfig.RSS_CLIENT_SORT_MEMORY_USE_THRESHOLD,
+        RssMRConfig.RSS_CLIENT_DEFAULT_SORT_MEMORY_USE_THRESHOLD);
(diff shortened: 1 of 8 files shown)


client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBufferManager.java
```diff
@@ -113,7 +114,7 @@ public SortWriteBufferManager(
       int numMaps,
       boolean isMemoryShuffleEnabled,
       int sendThreadNum,
-      double sendThreshold) {
+      double sendThreshold, long maxBufferSize) {
```
colinmjj, line 117: The new parameter should be start with new line.

client-mr/src/main/java/org/apache/hadoop/mapred/RssMapOutputCollector.java
```diff
@@ -121,7 +130,8 @@ public void init(Context context) throws IOException, ClassNotFoundException {
         numMaps,
         isMemoryShuffleEnabled(storageType),
         sendThreadNum,
-        sendThreshold);
+        sendThreshold,
+        1024000L);
```
colinmjj, line 134: it should be configured instead of 1024000L

jerqi: sorry, I'll fix this.

client-mr/src/main/java/org/apache/hadoop/mapred/SortWriteBufferManager.java
jerqi: OK

PR #169, opened 2022-06-09
[Improvement] Reduce the recomputation caused by bad node

jerqi:
As we know, when MRAppMaster find the node is a bad node,  and the node have execute some map tasks, MRAppMaster
 will recompute them, but it's not necessary for RSS, Because RSS don't store any shuffle data in those nodes. So we don't trigger any recomputation caused by bad node.

Reduce the recomputation, and recomputation will cause reduce fail because the loss of event.

No

Manual test in our cluster


client-mr/src/main/java/org/apache/hadoop/mapreduce/v2/app/RssMRAppMaster.java
```diff
@@ -206,22 +298,79 @@ static void writeExtraConf(JobConf conf, JobConf extraConf) {
...
-  // After we modify some configurations, we should update configuration for Application
-  // So we update the modify the configuration in the HDFS and local disk. It's a little
-  // tricky to delete the local configuration. But we hope guarantee the integrity for file.
-  // We choose to delete it and override with the new configuration in the HDFS.
-  static void updateConf(JobConf conf, Path jobConfFile) {
-    try {
-      File newFile = new File(MRJobConfig.JOB_CONF_FILE + ".bak");
-      try (FileOutputStream out = new FileOutputStream(newFile)) {
-          conf.writeXml(out);
+  private final class RssContainerAllocatorRouter extends AbstractService implements ContainerAllocator, RMHeartbeatHandler {
```
frankliee, line 301: Add more comment about how to close re-compute

colinmjj (approved):

PR #173, opened 2022-06-13
[Bugfix] Add retry mechanism when reduce merger runs out of memory

frankliee:
Add retry mechanism in reduce, which could avoid block missing when merger runs out of memory.

When merger runs out of memory, it will reserve null mapout. 
Without this PR, some fetched blocks will leak under this situation.  

No.

New UT is added to simulate merger that runs out of memory to reserve null mapout.


client-mr/src/main/java/org/apache/hadoop/mapreduce/task/reduce/RssFetcher.java
```diff
@@ -130,23 +136,26 @@ public void fetchAllRssBlocks() throws IOException, InterruptedException {
...
-      unCompressionLength += compressedBlock.getUncompressLength();
-      long decompressDuration = System.currentTimeMillis() - startDecompress;
-      decompressTime += decompressDuration;
+      if (!retry) {
+        final long startDecompress = System.currentTimeMillis();
+        uncompressedData = RssShuffleUtils.decompressData(
+          compressedData, compressedBlock.getUncompressLength(), false).array();
+        unCompressionLength += compressedBlock.getUncompressLength();
+        long decompressDuration = System.currentTimeMillis() - startDecompress;
+        decompressTime += decompressDuration;
```
jerqi, line 157: retry need sleep.

jerqi (dismissed):

client-mr/src/main/java/org/apache/hadoop/mapreduce/task/reduce/RssFetcher.java
frankliee: There is already a "sleep".
If it returns in this round, it will be blocked by `merger.waitForResource()`, which reduces the amount of usedMemory.

client-mr/src/main/java/org/apache/hadoop/mapreduce/task/reduce/RssFetcher.java
```diff
@@ -129,25 +133,30 @@ public void fetchAllRssBlocks() throws IOException, InterruptedException {
...
+      }
+      long fetchDuration = System.currentTimeMillis() - startFetch;
+      readTime += fetchDuration;
     }
-    long fetchDuration = System.currentTimeMillis() - startFetch;
-    readTime += fetchDuration;
 
     // uncompress the block
-    if (compressedData != null) {
+    if (!retry && compressedData != null) {
```
jerqi, line 150: better name

client-mr/src/main/java/org/apache/hadoop/mapreduce/task/reduce/RssFetcher.java
```diff
@@ -129,25 +133,30 @@ public void fetchAllRssBlocks() throws IOException, InterruptedException {
...
       final long startDecompress = System.currentTimeMillis();
-      byte[] uncompressedData = RssShuffleUtils.decompressData(
+      uncompressedData = RssShuffleUtils.decompressData(
```
jerqi, line 152: add wait time field.

client-mr/src/main/java/org/apache/hadoop/mapreduce/task/reduce/RssFetcher.java
frankliee: => `hasPendingData`

client-mr/src/main/java/org/apache/hadoop/mapreduce/task/reduce/RssFetcher.java
frankliee: Done

client-mr/src/main/java/org/apache/hadoop/mapreduce/task/reduce/RssFetcher.java
```diff
@@ -201,11 +189,55 @@ public void copyFromRssServer() throws IOException {
...
       LOG.info("reduce task " + reduceId.toString() + " cost " + readTime + " ms to fetch and "
         + decompressTime + " ms to decompress with unCompressionLength["
-        + unCompressionLength + "] and " + serializeTime + " ms to serialize");
+        + unCompressionLength + "] and " + serializeTime + " ms to serialize and "
+        + waitTime + " ms to wait resource");
       stopFetch();
     }
   }
 
+  private boolean issueMapoutMerge() throws IOException {
```
jerqi, line 198: better name?

frankliee:
Do you have any suggestions ? @colinmjj

jerqi:
LGTM

colinmjj (approved):

PR #177, opened 2022-06-16
[Improvement] Only initialize the heartbeat thread pool when the role is driver

zuston:
Only initialize the heartbeat thread pool when the role is driver

Just to avoid creating extra thread pool.

No

No


jerqi:
The spark2 also have similar problems. Could you modify the spark2 at the same time?

zuston:
Updated @jerqi .

I'm confused that when heartbeat failed, it will do nothing. Right?

colinmjj:
> Updated @jerqi .
> 
> I'm confused that when heartbeat failed, it will do nothing. Right?

Heartbeat is to notify shuffle server app is alive, and every rpc also will do the notify. According to current implementation, ignore the failed rpc.

jerqi:
If shuffle server haven't received any heartbeat of application, shuffle server will delete the shuffle data of the application.

jerqi (approved):
LGTM, Thanks for your contribution.

PR #180, opened 2022-06-16
Avoid using the default forkjoin pool by parallelStream directly

zuston:
As we know that parallelStream will use the default forkjoin pool in entire jvm. To avoid it, use the custom pool and allow to specify the pool size.

use separate forkjoin pool to send shuffle data

Yes, introduce the configuration to control the size of forkjoinpool.
mapreduce.rss.client.data.transfer.pool.size for MapReduce
spark.rss.client.data.transfer.pool.size for Spark

GA passed.

### DIFF
--- a/client/src/main/java/com/tencent/rss/client/impl/ShuffleWriteClientImpl.java
+++ b/client/src/main/java/com/tencent/rss/client/impl/ShuffleWriteClientImpl.java
@@ -24,6 +24,7 @@ import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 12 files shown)


zuston:
And although we can specify the rss.client.send.pool.size, but the latency may depends on the default forkjoin size in current implementation. Right?

zuston:
@frankliee Do u mind reviewing this?

jerqi:
Actually I prefer thread pool, I find some problems when we use fork join pool. You can know more information about it to see  #69 .

zuston:
Thanks for pointing this @jerqi .

And let me guess what it happened like #69. Maybe firestorm use the parallelstream in multiple methods, and we know the default forkjoin pool that only one in entire JVM will be used. So if some methods occupy the fork join pool's thread, other methods will be blocked, and the pool size number is limited to the machine cpu cores.

So i think this is not the bug of parallelstream, just the incorrect usage.

Please let me know what u think

jerqi:
wecom-temp-900cc0ca86634642056c5ace1c33ecb0
I find the jstack at that time, maybe, you are right.

zuston:
Just guess. @jerqi . But only this jstack screenshot can't prove it.
 Pity.

zuston:
Hi. So do i need to use threadpool instead of parallelstream of forkjoin pool? @jerqi @colinmjj

jerqi:
I think parallelstream of forkjoin pool is ok. What do you think?@colinmjj

colinmjj:
parallelstream of forkjoin pool is ok for me, @zuston @jerqi

client/src/main/java/com/tencent/rss/client/impl/ShuffleWriteClientImpl.java
```diff
@@ -125,17 +131,17 @@ private boolean sendShuffleDataAsync(
...
             LOG.warn("Send: " + serverToBlockIds.get(ssi).size() + " blocks to [" + ssi.getId()
-              + "] failed with statusCode[" + response.getStatusCode() + "], ");
+                    + "] failed with statusCode[" + response.getStatusCode() + "], ");
           }
         } catch (Exception e) {
           isAllServersSuccess.set(false);
           LOG.warn("Send: " + serverToBlockIds.get(ssi).size() + " blocks to [" + ssi.getId() + "] failed.", e);
         }
-      });
+      })).join();
```
jerqi, line 144: If there are two threads to call the same forkJoinPool's join method,  is it `join method` safe?

zuston: Yes. The join will only wait its current thread's tasks.

jerqi: OK, got it.

jerqi (approved):
LGTM, thanks for your contribution.

PR #183, opened 2022-06-20
[Minor] Prevent returning incomplete shuffle blocks

zuston:
The problem only occurs when `wb.getMemoryUsed() > bufferSize` and `usedBytes.get() - inSendListBytes.get() > spillSize`. 

Although it doesn't seem to happen because of the logic, i think it should be optimized by using `addAll` to avoid confusion.

Better code

No

No need.

### DIFF
--- a/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
+++ b/client-spark/common/src/main/java/org/apache/spark/shuffle/writer/WriteBufferManager.java
@@ -146,7 +146,7 @@ public class WriteBufferManager extends MemoryConsumer {
 
     // check buffer size > spill threshold
     if (usedBytes.get() - inSendListBytes.get() > spillSize) {
-      result = clear();
+      result.addAll(clear());
     }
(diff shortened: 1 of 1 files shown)


zuston:
@jerqi Minor fix.

jerqi:
Could you add the ut for this fix?

zuston:
I think this bug will not happen, so current test cases are enough.

The condition of  `wb.getMemoryUsed() > bufferSize` and `usedBytes.get() - inSendListBytes.get() > spillSize` are mutually exclusive.

@jerqi If i'm wrong, please point it. Thanks.

jerqi:
> I think this bug will not happen, so current test cases are enough.
> 
> The condition of `wb.getMemoryUsed() > bufferSize` and `usedBytes.get() - inSendListBytes.get() > spillSize` are mutually exclusive.
> 
> @jerqi If i'm wrong, please point it. Thanks.

Yes, you're right. It will not happen. Please rebase master. If CI passed, I will merge it.

jerqi (approved):
LGTM, Thanks for your contribution.

PR #184, opened 2022-06-20
[Improvement] Disable flaky test

jerqi:
Disable flaky test

The test is not important, but it's very flaky. We disable this test temporarily. We will fix this flaky test later.

No

No need.


colinmjj (approved):
LGTM, it should be fixed later

PR #186, opened 2022-06-22
[Bugfix] Fix spark2 executor stop NPE problem

jerqi:
We need to judge heartbeatExecutorService whether is null when we will stop it.

#177 pr introduce this problem, when we run Spark applications on our cluster, the executor will throw NPE when method `stop` is called.

No

Manual test


colinmjj (approved):
LGTM

jerqi:
Later we will add Yarn MiniCluster UT.

PR #196, opened 2022-06-23
[Bugfix] [0.5] Fix MR don't have remote storage information when we use dynamic conf and MEMORY_LOCALE_HDFS storageType

jerqi:
backport 0.5.0

We should acquire the storageType from extraConf. 

If we don't have this patch, MR don't work when we use dynamic conf and MEMORY_LOCALE_HDFS storageType.

No

Manual test


colinmjj (approved):
LGTM

PR #200, opened 2022-06-24
Support using remote fs path to specify the excludeNodesFilePath

zuston:
Support using remote fs path to specify the excludeNodesFilePath

When existing two coordinators serving for online, we hope they can read the consistent exclude nodes file insteading of using the localfile syncing manually.

Yes. It's an incompatible change. 

When the default fs is HDFS in the core-site.xml, and the excludeFilePath is specified to "/user/xxxxx" in coordinator server.
After applied this patch, filesystem will be initialized to remote HDFS due to lacking scheme.

Unit tests.

### DIFF
--- a/coordinator/src/main/java/com/tencent/rss/coordinator/ClusterManagerFactory.java
+++ b/coordinator/src/main/java/com/tencent/rss/coordinator/ClusterManagerFactory.java
@@ -18,15 +18,19 @@
 
 package com.tencent.rss.coordinator;
 
+import org.apache.hadoop.conf.Configuration;
+
 public class ClusterManagerFactory {
 
   CoordinatorConf conf;
+  Configuration hadoopConf;
 
-  public ClusterManagerFactory(CoordinatorConf conf) {
+  public ClusterManagerFactory(CoordinatorConf conf, Configuration hadoopConf) {
     this.conf = conf;
+    this.hadoopConf = hadoopConf;
   }
 
-  public ClusterManager getClusterManager() {
-    return new SimpleClusterManager(conf);
+  public ClusterManager getClusterManager() throws Exception {
+    return new SimpleClusterManager(conf, hadoopConf);
   }
 }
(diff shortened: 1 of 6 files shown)


zuston:
@jerqi @colinmjj Could u help review?

jerqi:
You seems to use the HDFS as configuration center. I think we should have abstraction `ExcludeNodesConfigProvider`.  We can provide `HdfsExcludeNodesConfigProvider` and `LocalExcludeNodesConfigProvider`. Users can choose one of them. Local file is convenient, and uses don't care consistence sometimes.

zuston:
As we know, Hadoop Filesystem also support reading local file. In this PR, users also can make excludeNodoFilePath as local file path. So the previous unit tests work. refer to: https://github.com/Tencent/Firestorm/blob/6bdf49e1a68131545a8385123da558be287a196f/coordinator/src/test/java/com/tencent/rss/coordinator/SimpleClusterManagerTest.java#L173

But as described in description, it have to specify local file path as `file:/xxxxxx ` ,when using hadoop filesystem reading local file and the default-fs  has been set in the core-site.xml

zuston:
> excludeNodesFile is a workaround solution to do the node management. I think we can have a better way to do it, eg, coordinator support command line/restApi for these management requirement.

As i know, the excludeNodesFile is designed to make shuffle server decommission gracefully. 

Maybe we can make the excludeNodesFile as the one kind of exclude nodes strategy and provide more strategy to exclude node like cli/rest api. Do u think so? @colinmjj 

If we want to preserver the excludeNodesFile, i think this PR is an improvement.

zuston:
Please let me know what u think. Thanks~ @jerqi @colinmjj

jerqi:
For the long term, we prefer to implementing a configuration center and removing the dependency of hadoop's jar. The class `ClientConfManager` in the Coordinator Server also use the method like this pr to manage its conf.  So I think it's ok.@colinmjj WDYT

zuston:
@colinmjj Gentle ping

colinmjj:
OK, we can add this and improve the solution in another PR

colinmjj (approved):
LGTM

zuston:
Thanks @jerqi @colinmjj

PR #204, opened 2022-06-28
[Improvement] Add RSS_IP environment variable support for K8S

jerqi:
Method `getHostIp` can acquire IP by environment variable.

For K8S, there are too many IPs, it's hard to decide which we should use. So we use the environment variable to tell RSS
to use which one.

NO

UT


colinmjj (approved):
LGTM

PR #205, opened 2022-06-29
Close coordinatorClients when DelegationRssShuffleManager stops

wForget:

Close coordinatorClients when DelegationRssShuffleManager stops.

The coordinatorClients in DelegationRssShuffleManager are never closed.

No

No

### DIFF
--- a/client-spark/spark3/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
+++ b/client-spark/spark3/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
@@ -248,6 +248,7 @@ public class DelegationRssShuffleManager implements ShuffleManager {
   @Override
   public void stop() {
     delegate.stop();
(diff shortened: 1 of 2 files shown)


client-spark/spark3/src/main/java/org/apache/spark/shuffle/DelegationRssShuffleManager.java
```diff
@@ -248,6 +248,7 @@ public boolean unregisterShuffle(int shuffleId) {
   @Override
   public void stop() {
     delegate.stop();
+    coordinatorClients.forEach(CoordinatorClient::close);
```
colinmjj, line 251: spark2 should have the same problem?

wForget: > spark2 should have the same problem?

You are right, I will fix it in spark2 as well.

jerqi (approved):
LGTM, thanks for your contribution.

PR #215, opened 2022-07-22
[Bugfix]Fix memory leak & Improve buffer management

summaryzb:
the behavior of buffer management and memory allocate

1.fix memory leak, case when adding a huge record to a partition which buffer is not initialized
2.improve buffer management, byte array will be fully used as  possible to  save memory

no

through the testing


roryqi:
Hi @summaryzb  , we're moving to new repo https://github.com/apache/incubator-uniffle now, please create a new PR against that repo, thanks!

summaryzb:
> Hi @summaryzb , we're moving to new repo https://github.com/apache/incubator-uniffle now, please create a new PR against that repo, thanks!

thanks for reminding

